PR #2, opened 2022-04-17
Fix uninitialized reads

argilo:
Valgrind reports a couple places where data is used before it is initialized:

```
==12481== Conditional jump or move depends on uninitialised value(s)
==12481==    at 0x11E2D5: setup_mcast (multicast.c:63)
==12481==    by 0x10C70D: loadconfig (main.c:349)
==12481==    by 0x10BB6A: main (main.c:155)
```
If `target` is null, then the uninitialized `iface` will be passed into `strlen`.
```
==12481== Conditional jump or move depends on uninitialised value(s)
==12481==    at 0x113092: alloc_demod (radio.c:63)
==12481==    by 0x10C7FF: loadconfig (main.c:372)
==12481==    by 0x10BB6A: main (main.c:155)
```
The `inuse` field is read even though `Demod_list` is not initialized.

I've fixed both problems here.

### DIFF
--- a/multicast.c
+++ b/multicast.c
@@ -58,12 +58,11 @@ int setup_mcast(char const * const target,struct sockaddr *sock,int const output
     memset(sock,0,sizeof(struct sockaddr_storage));
   }
   char iface[1024];
+  iface[0] = '\0';
   if(target)
     resolve_mcast(target,sock,DEFAULT_RTP_PORT+offset,iface,sizeof(iface));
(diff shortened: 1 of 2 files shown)


ka9q (approved):
I'm surprised gcc didn't flag these, it seems to flag similar errors I've made elsewhere. Thanks.

argilo:
Closing, since you already manually applied these changes.

PR #5, opened 2022-04-17
Add .gitignore file

argilo:
Ignoring build artifacts will make it easier to work with git.


argilo:
It's a bit annoying that the executables have to be listed individually. Perhaps placing them in a directory (e.g. `bin`) would make this easier.

ka9q (approved):

ka9q:
How do I automatically build them in a bin directory? If I can do that 
easily with classic make, sure.

I *strongly* resist the use of a meta-build tool such as cmake. Almost 
every time I've had to use it in somebody else's package I run into 
problems that require an hour of digging into its bowels to figure out 
some trivial error.

ka9q:
Keep in mind I am still only barely literate with git, and right now I 
really don't want to have to learn more of it than I absolutely have to.

argilo:
> How do I automatically build them in a bin directory? If I can do that easily with classic make, sure.

I had a look, and it doesn't seem to be easy. So I think things are fine as they are. Keeping a list of executables in `.gitignore` isn't so bad.

PR #8, opened 2022-04-21
Update the list of required Debian packages

argilo:
The information in `debian-deps.txt` is now duplicated in `INSTALL.txt`, so we may as well remove the older file.

The `build-essential` package is needed to get a C compiler and `make`.

The `libudev-dev` package does not seem to be required; I built ka9q-radio just fine without it.

The dependence on `libattr-dev` was removed in 6fa9c1a85f9ec244a59ead7d982c3ed166f86a46.


ka9q:
Oh, good to know about libattr-dev! I just removed it from my Debian x86 
and my Raspberry Pis, and sure enough ka9q-radio still built cleanly. I 
wonder how long ago the attr stuff was put into libc.

I *love* getting rid of dependencies...

PR #9, opened 2022-04-22
Ignore Makefile symlink

argilo:
The installation instructions suggest creating a `Makefile` symlink, so it makes sense to include that in the list of files that git should ignore.


ka9q (approved):
Good idea. I'm still a novice with git.

ka9q:
Change made (I think).

PR #11, opened 2022-04-22
Check the return value of asprintf

argilo:
gcc warns about a few places where the return value of `asprintf` is not checked. I've added checks here, and no warnings remain.


ka9q (approved):
Thanks. Wonder why it only complains about asprintf. I ignore return values from lots of functions, as does everyone else. True, nasty things could happen if the function fails and I try to use the output pointer.

PR #14, opened 2022-05-10
Fix signed integer overflows

argilo:
GCC's Undefined Behaviour Sanitizer (`-fsanitize=undefined`) identifies several signed integer overflows when airspyd is started:
```
avahi.c:289:42: runtime error: left shift of 239 by 24 places cannot be represented in type 'int'
airspyd.c:785:30: runtime error: left shift of 1770000000 by 1 places cannot be represented in type 'int'
multicast.c:284:26: runtime error: left shift of 2 by 30 places cannot be represented in type 'int'
```
I've corrected those here by switching to unsigned integer types. I also corrected the same issue in rtlsdr.

### DIFF
--- a/airspyd.c
+++ b/airspyd.c
@@ -781,8 +781,8 @@ int rx_callback(airspy_transfer *transfer){
 // Not easy to fix without knowing the calibration parameters.
 // Best workaround is a GPSDO, which disables the correction
 double true_freq(uint64_t freq_hz){
-  int const VCO_MIN=1770000000u; // 1.77 GHz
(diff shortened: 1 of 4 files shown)


ka9q (approved):
Looks good. I try to make a point of using unsigned in these situations, but I sometimes miss them. Here the values are actually interpreted as unsigned so code functions correctly, but it's best to be consistent.

ka9q:
I made some slight tweaks to your fixes. Rather than cast a hex constant to unsigned, I just appended "u".

PR #15, opened 2022-05-10
Add GitHub Actions CI

argilo:
In case you'd like to add CI to the project, I've created a basic GitHub Actions workflow here. It verifies that the build and install succeeds when using each of the three Makefile variants.

If desired, additional test steps could be added.

The output of a sample run on my fork can be seen here: https://github.com/argilo/ka9q-radio/actions/runs/2301026546

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,82 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  ubuntu-build:
+    name: Ubuntu CI
+    strategy:
+      matrix:
+        makefile: [Makefile.linux, Makefile.debug]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y \
+            libusb-1.0-0-dev \
+            libncurses5-dev \
+            libfftw3-dev \
+            libbsd-dev \
+            libhackrf-dev \
+            libopus-dev \
+            libairspy-dev \
+            libairspyhf-dev \
+            librtlsdr-dev \
+            libiniparser-dev \
+            libavahi-client-dev \
+            portaudio19-dev
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup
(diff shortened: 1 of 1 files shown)


ka9q:
I don't know what that means, but if it doesn't interfere with anything I'm doing I'm OK with it.

ka9q:
I notice you're using 'brew' on MacOS. I use 'port'. Is there any meaningful difference?

argilo:
Basically it means that every time a commit is pushed to GitHub, or a pull request is opened, GitHub will spin up containers and build the project, verifying that no new compile errors have been introduced. GitHub will show a green check mark next to the commit if the builds succeed. It shouldn't interfere with your work, but may alert you to bugs.

I used brew because it is available by default on GitHub's build machines, while port is not.

ka9q:
I didn't even know that Github could do this, but as long as it doesn't interfere with anything I have no objection. And if their compilers spot problems that mine don't, this could be helpful.

ka9q:
I've been doing frequent commits of small changes, would this create a problem?

argilo:
That shouldn't be a problem. If a lot of commits are pushed around the same time, GitHub simply queues up the builds. In that event, it will just take a bit longer to see whether a given commit passed or failed.

argilo:
At the moment, the execution time is approximately 1 minute for the Linux builds, and 5 minutes for the MacOS build. (MacOS takes a bit longer because avahi must be built from source; it's not available as a brew package.)

ka9q:
So is there any downside to my accepting this pull request?

argilo:
I don't see any downside.

If it doesn't work out and you want to get rid of it, all you have to do is delete the `.github` folder.

argilo:
At the moment, this just verifies that `make` and `sudo make install` execute without errors on Ubuntu and macOS. It could certainly be extended to test other things. GitHub actions runs arbitrary shell commands, so we could ask it to run the executables once they're built, run a test suite, or even produce a pre-packaged version of the software for download.

ka9q:
I guess we could write some test programs. Send commands to a radiod or airspyd instance, see if they get the expected result. Unit tests are a good idea anyway.

PR #16, opened 2022-05-10
Terminate longopts arrays

argilo:
The "tune" and "setfilt" programs currently crash if an invalid command line argument is given (e.g. `tune --foo` or `setfilt --foo`). This happens because the `longopts` arrays passed into `getopt_long` are not terminated. I've added the missing terminations here.


ka9q (approved):
Thanks. Those programs were quick hacks and I haven't used them much. Glad to know someone else is watching more closely!

PR #23, opened 2022-08-30
Add glib dependency to macOS build

### LINKED ISSUE #21: Automatic test builds for MacOS are failing
The automatic build tests for MacOS are regularly failing now. They seem to fail while still constructing the test environment, even before it gets to my code. I don't understand how this system works; can anybody easily figure out what's going on?

argilo:
This package is required to build avahi.

Fixes #21.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
         run: |
           brew update
           brew install \
+            glib \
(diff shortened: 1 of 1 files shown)


ka9q:
Thanks for fixing this! I had no problems building ka9q-radio on my Macbook, so I guess I already had that dependency.

PR #26, opened 2022-10-16
fix build on Arch Linux

pinkavaj:
The header iniparser.h on Arch linux is not in subdirectory, this is a bit hacky approach bud should allow build on both Ubuntu and Arch Linux.

The other options to consider is to add more aleborate build system (like CMake), I can try to do that if desired.


ka9q (approved):
Grr. It's this kind of gratuitous portability problem that drives me nuts.

ka9q:
Only this breaks things on MacOS! Grr....

ka9q:
Maybe it *is* time for cmake...

pinkavaj:
I can try to fix the Mac issue, but I does not have one, so I need the error description.

PR #28, opened 2022-10-16
rtlsdr: increase the block size to 1298

pinkavaj:
I'm completely ignorant why 960 was used and there is no documentation, any special reason?

This reduces the number of packets to send (and thus switching overhead), it is also 2*512+256 -> it is better aligned with USB transfer size.

Btw, when jumbo frames are used, then the 8192 block size works like a charm, would make sense to make this configurable?

### DIFF
--- a/rtlsdr.c
+++ b/rtlsdr.c
@@ -40,7 +40,7 @@
 
 // Internal clock is 28.8 MHz, and 1.8 MHz * 16 = 28.8 MHz
 #define DEFAULT_SAMPRATE (1800000)
-#define DEFAULT_BLOCKSIZE (960)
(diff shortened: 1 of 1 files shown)


ka9q:
The IQ packet size stuff in the various front end handlers is rather ad hoc. Back when I was only using the Funcube dongle, its low sample rate meant that a large packet added delay. I was also trying to use values that were multiples of the block sizes I was then using in what is now radiod. I don't use the rtl_sdr much, so that code still needs a lot of cleanup. So yes, larger packets make more sense for devices with higher sample rates, especially when jumbo frames are used and especially^2 when very large frames are used for loopback (e.g., IP TTL=0).

ka9q (approved):

PR #30, opened 2022-10-17
add function to get distribution file path

pinkavaj:
This allows run the application from either the build directory or from installation directory. This is usefull for mostly development and debugging.


ka9q (approved):
Good idea. I've always thought my handling of library file paths needed improvement.

PR #31, opened 2022-10-20
decode_status: fix ssrc decoding

pinkavaj:
We are connecting to output, so the SSRC is labeled as output althought from our perspective it is input


ka9q (commented):
Hmm, I see your point. Yet 'control' is displaying the correct SSRC; how could this work? I'm investigating...

ka9q:
Oh, I see why. It still works because 'control' is getting the front end's output SSRC via the 'input' SSRC field from radiod. And even though radiod also uses decode_fe_status() it *also* gets the correct SSRC when it calls rtp_process(). So yeah, it's a bug, but one that didn't cause any problems. They can be hard to find!

PR #32, opened 2022-11-14
Rename rtlsdr to rtlsdrd to match other daemon naming

pinkavaj:


ka9q:
Thanks. The rtlsdr(d) program is definitely less mature than the other front end programs.

Makefile.debug
```diff
@@ -106,7 +106,7 @@ control: control.o modes.o bandplan.o decode_status.o libradio.a
 cwd: cwd.o morse.o libradio.a
 	$(CC) $(LDOPTS) -o $@ $^ -lbsd -lpthread -lm
 
-rtlsdr: rtlsdr.o libradio.a
+rtlsdr: rtlsdrd.o libradio.a
```
argilo, line 109: As reported by CI, this change broke compilation on Linux because this rule was not renamed to `rtlsdrd`.

argilo: I've fixed it in #34.

PR #33, opened 2022-11-14
Build using CMake

pinkavaj:
Replace plain `make` build by `CMake`. The goal is to:

* Make the build system more platform independent
* Check for presence of libraries and build only relevant modules

TODO:
- [ ] Simplify the initial CMake file
- [ ] Add necessary documentation to ensure easy build for every newcommer
- [ ] Make the builds which depend on specific libraries optional (eg, rtlsdr, funcube ...)
- [ ] Ensure all the features and optimisations from old Makefile are ported correctly.
- [x] Test on Linux
- [ ] Test on MacOsX
- [ ] Test on other system?


ka9q:
Thank you! I've never written for cmake myself.

pinkavaj:
@ka9q Ok, the initial version is ready, there is still some tweaks to do, but feedback is really welcome (from anybody of course).

ka9q:
Sorry for letting this sit for so long. I'm still unfamiliar with both cmake and with handling git pull requests. Do you still want me to merge this, or do you want to look at it again?

ka9q:
I've cleaned up the existing makefiles quite a bit, so I'm not sure how important it still is to switch to cmake right now. Comments?

ka9q:
This has been sitting here for a while and I've probably made too many changes to my own makefile to merge this easily. Also, my newer makefiles can handle build options. So I'm closing this for now but we can revisit cmake in the future if you're still interested.

PR #34, opened 2022-11-19
Fix rule name in Makefile.debug

argilo:
This fixes the rule name that was missed in #32.


pinkavaj:
The same error is in `README.md` file and possibly in `rtlsdrd.c` header.

argilo:
I updated the readme as well. I'll leave the header, since I believe it's automatically generated by RCS.

pinkavaj (approved):

ka9q (approved):
Thanks. Makefile.debug is generated manually, and I simply forgot to update it.

PR #41, opened 2023-10-10
Restore CI

argilo:
I noticed that CI was removed due to consistent build failures on macOS. The problem was simply that CI called for macOS 11, which is no longer supported by Homebrew. I've switched the build to request the latest stable macOS release, which fixes the problem and should avoid future trouble the next time a macOS version goes out of support. I did the same for Ubuntu, so that Linux CI continues to run on a supported platform as well.


ka9q:
I really don't see much use to these "workflows". All they ever do is fill my mailbox with error messages. I know how to see if my code compiles. Do they really serve a purpose?

argilo:
Someone who submits a pull request may not have compiled on both platforms (or may have forgotten to update `Makefile.debug`), so the CI checks would alert them to the problem.

The checks could become more useful if unit tests were added to ka9q-radio someday. If there are any manual checks you perform today beyond verifying compilation on Linux and Mac, those could perhaps be added to CI.

If you don't want the emails, they can be disabled in https://github.com/settings/notifications. See the Systems -> Action section near the bottom.

argilo:
CI seems to be failing for good reason at the moment. Compilation is indeed broken on Linux in the same way as the workflow reports.

argilo:
Workflow output:

https://github.com/ka9q/ka9q-radio/actions/runs/6556442875/job/17806448294

Building locally, I also get `make: *** No rule to make target 'wspr-decoded', needed by 'all'.  Stop.`

argilo:
It looks like the change in e8190a55304f093727cd3ca48488c7f68f58c6e4 removed the `wspr-decoded` target, but it was not removed from `EXECS`.

An automated compile check is good at catching these sorts of bugs.

PR #42, opened 2023-10-18
Remove references to wspr-decoded

argilo:
wspr-decoded was removed in e8190a55304f093727cd3ca48488c7f68f58c6e4, but it is still referenced in the Linux makefiles, causing the build to fail (unless a stale wspr-decoded executable happens to be around from a previous compile). Here I've removed the references, after which CI should pass again.

I also removed wspr-decodec from .gitignore, where it should no longer be required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,7 +41,6 @@ show-sig
 stereod
 tune
 wd-record
-wspr-decoded
 old
 RCS
 push.sh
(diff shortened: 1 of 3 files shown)


argilo:
I'll close this off, since I see you already made the fix.

`wspr-decoded` is still present in `.gitignore`, but it's not causing any trouble there.

PR #50, opened 2024-03-30
Mod pcmrecord to include center frequency for SDR Console

scottnewell:
SDR Console loads the IQ .wav files generated by pcmrecord, but it assumes they are at a 10 MHz center frequency. This patch adds a (mostly undocumented?) 'auxi' chunk to the .wav file to specify the center frequency (and stop/stop times). This was to support VE7DXR Nick Hall-Patch's Hamsci MW recording project (https://hamsci.org/mw-recordings/).


ka9q (approved):
Looks good, thanks!

PR #51, opened 2024-03-30
Modify pcmrecord length limit to consider number of channels

scottnewell:
pcmrecord length limit and diag output assumes single channel files.


ka9q:
Thanks!

PR #55, opened 2024-05-11
Fix macOS CI build

argilo:
The macOS CI build began failing when macOS 14 became the default on Github Actions. I made the following changes to fix that:

* Use `$(brew --prefix)` to determine where Homebrew has installed packages, instead of using hard-coded paths.
* Add missing `#include` directives.

Also, I updated the checkout actions to version 4, which fixes deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
             libavahi-client-dev \
             portaudio19-dev
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup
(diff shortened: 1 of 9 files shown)


argilo:
It worked at the time, but it looks like there's now an issue with the iniparser package on macOS. I'll have a look into that tonight.

ka9q:
What was the problem? I used iniparser because it seemed simple, but maybe it isn't? Are there any other parsers for the same config file format?

argilo:
The problem was not with iniparser itself, but that Homebrew had not packaged it correctly. That's now fixed, so I had to remove my workaround. See #60.

PR #63, opened 2024-11-12
Fix RTL-SDR gain settings

argilo:
Here I have corrected a few problems that prevent RTL-SDR gains from being set:

1. The `rtlsdr_set_direct_sampling` call has the side effect of resetting the gain settings, so it must be called before setting the gain. (For good measure, I moved the `rtlsdr_set_offset_tuning` call earlier as well.)
2. The `rtlsdr_set_tuner_gain` call has the side effect of turning off AGC, so it must not be called after AGC is enabled.
3. The `rtlsdr_set_tuner_gain_mode` argument was reversed. According to the documentation, 0 enables AGC, and 1 switches to manual gain mode.

Since ka9q-radio's `gain =` configuration setting was not working anyway, I took the liberty of changing it to be a floating-point value in dB. That brings it in line with the standard RTL-SDR utilities, which multiply the gain argument by 10 and convert it to an integer before passing it into `rtlsdr_set_tuner_gain`, which takes the gain in tenths of a dB.

### DIFF
--- a/rtlsdr.c
+++ b/rtlsdr.c
@@ -147,6 +147,8 @@ int rtlsdr_setup(struct frontend *frontend,dictionary *dictionary,char const *se
     fprintf(stderr,"\n");
 
   }
+  rtlsdr_set_direct_sampling(sdr->device, 0); // That's for HF
+  rtlsdr_set_offset_tuning(sdr->device,0); // Leave the DC spike for now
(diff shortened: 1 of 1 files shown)


ka9q:
I don't use the RTL-SDR so I'll take it that your fixes work!

PR #68, opened 2024-12-01
Add libogg dependency to CI builds

argilo:
ka9q-radio now depends on libogg, so it needs to be added to the CI workflows. I've done that here, and the CI runs now pass.


ka9q:
Thanks! Yes, I'm working on pcmrecord to record Opus streams to ogg Opus files (suffix .opus). Still need to resequence out-of-order packets since they're of unknown size.

PR #71, opened 2024-12-07
Round bin powers to two decimal places

argilo:
The "powers" utility currently rounds bin powers to one decimal place. When using a long integration period (like Radiosonde Auto-RX does), quantization becomes readily visible:

Screenshot from 2024-12-07 11-12-16

I think it would make sense to increase the precision to two decimal places. This would also match the behaviour of rtl_power.


ka9q (approved):
No problem.

Issue #62, opened 2024-11-05
Problems with "powers" utility


argilo:
I'm helping with development and testing of radiosonde_auto_rx, and noticed some problems with the "powers" utility:

1. ~~If the specified SSRC does not exist yet, powers creates it, but the first returned row is garbage:~~
```
$ powers radiosonde.local -f 403025000.0 -w 10000 -b 10 -i 1 -s 1002 -c 2
Tue 05 Nov 2024 17:03:54.946267 UTC, 402974976, 403064976, 10000, 10, -100.0, -100.0, -100.0, -100.0, -100.0, -439.5, -100.0, -100.0, -100.0, -100.0,
Tue 05 Nov 2024 17:03:55.963905 UTC, 402974976, 403064976, 10000, 10, 28.0, 28.3, 28.5, 28.4, 27.8, 27.5, 27.5, 27.3, 27.3, 27.3,
```
2. The first row of data is requested and displayed immediately, regardless of the integration interval. As a result, it may contain data that was integrated over a longer or shorter period than requested.
3. 500ms after each row of data is requested, an output timer fires, which seems to reset the spectrum bin data. This shortens the requested integration time by 500ms.
4. At one point, it seemed like a timer was firing every 500ms and resetting the bin data, thus causing the integration interval to be very short all the time. (I haven't been able to reproduce the conditions that caused this yet.)
5. ~~The time format (`Tue 05 Nov 2024 17:03:54.946267 UTC`) seems to be intended for human consumption. `2024-11-05T17:03:54.946267Z` might be more appropriate.~~
6. The requested bin bandwidth is rounded to the nearest multiple of the FFT bin size, but the bin bandwidth, lower & upper frequencies reported by `powers` are not adjusted to take this rounding into account.

argilo:
I had a peek at the code, and it seems like (4) could occur if `silent` was false for the SPECT_DEMOD channel, which would cause the output timer to restart. I'm not sure how to get it into that state, though.

I also noticed that the code that responds to commands calls `send_radio_status` twice, and each of those calls `encode_radio_status`, which clears `chan->spectrum.bin_data`. Thus the second response (which is probably not used by powers) has all bins set to zero. `chan->status.blocks_since_poll` is not reset until the single `reset_radio_status` call which occurs later. Perhaps the bin data should be cleared there instead?

argilo:
[...]

PR #72, opened 2024-12-07
Use ISO 8601 format for time in powers output

argilo:
This addresses issue 5 from #62.


ka9q (commented):
Looks OK to me. BTW, pcmrecord uses something very similar to ISO 8601 format in its file names.

PR #74, opened 2024-12-12
Add 'fact' chunk to pcmrecord

scottnewell:
Adding the 'fact' chunk to the wav file should silence the "WARN wav: wave header missing extended part of fmt chunk" messages from sox. I'm not certain if it's needed for the integer wav files, but I don't think it'll hurt.


ka9q (approved):
I'm not using these headers so I'll take your word for it that this works...

PR #75, opened 2024-12-13
Update SDRplay Support

ke5gdb:
This PR is to merge the SDRplay support by K4VZ into the main branch. Compiling for SDRplay support is done by means of `make -f Makefile.sdrplay` so it can remain separate from the other mainline needs.


ka9q (approved):
I'll assume this works. I have some SDRPlays myself but have never tested them. I will also look closely at your suggested changes for loopback support, that would also solve the problem of running an isolated system not on a network.

PR #76, opened 2024-12-15
Add metadump option to read from stdin

scottnewell:


ka9q (approved):
I assume this is intended only to read and decode a single status packet from standard input? It looks like it, and there's no way to delimit multiple packets from a byte stream.

PR #78, opened 2024-12-18
Verify preset pointer is valid, handle TLV types >= 128 bytes

scottnewell:
This patch seems to have stopped the segfaults when tuning a spec demod stream to 0 Hz. Stole the optlen code from dump.c. Should tune link the dump_metadata() function in dump.c instead of repeating some of the TLV parse code in tune.c? That's above my pay grade.


ka9q:
I wrote tune.c as a quick hack, it probably needs to be rewritten to be much more complete and featureful.

PR #80, opened 2024-12-19
Try and avoid segfaulting in filter.c?

scottnewell:
WAG, but it seems to have stopped the segfaults. Not sure if it breaks anything else, so please review this!


ka9q:
Already did the same thing!

PR #83, opened 2024-12-31
Fix aviation config

argilo:
The aviation config file references a mode that no longer exists (`cam`). It looks like `amsq` is intended for aviation AM signals, so I've switched to that. I also removed the `samprate` setting, which I believe is unnecessary, and the `squelch` setting, which doesn't seem to exist.


ka9q (approved):
Looks OK, I haven't used this one in a while. Thanks for catching it.

PR #84, opened 2024-12-31
Fail if default status and data channels share the same name

scottnewell:
The sample config file in ka9q-web has this error, resulting in a webpage with a weird flickering spectrum display.


ka9q (approved):
Looks ok, but it should probably compare the resolved socket addresses (i.e, the IP addresses corresponding to the names). Yes, the command and data multicast groups are assumed to be distinct. I haven't actually tried making them the same, but I suspect my practice of resetting spectral data every time it is read is causing the problem. I've been talking about changing it to exponential smoothing, I should try that soon.

scottnewell:
I wasn't at all sure the right way to go, just wanted to add _something_ defensive in case anyone else got bit by the same ka9q-web bug.

The flickering web display from the duplicate names did not look like the spectrum display 'bouncing' we've discussed in the past. I think this severe flicker was a different issue caused entirely by the ka9q-web server.

PR #86, opened 2025-01-05
Update control help file

scottnewell:
Not sure if my makefile change is a good idea or not. Feel free to ignore.


ka9q:
I think I've already handled this.

scottnewell:
Did you notice the makefile patch? I didn't realize the help.txt wasn't updated during make install, so mine was terribly far out of date. I get that you don't want to overwrite an existing local file in case there were local changes; not sure if that's a concern with the control and monitor help files. At least I know to do it manually now!

PR #87, opened 2025-01-05
Fix pcmrecord wav file auxi chunk start timestamp

scottnewell:
If the "lengthlimit" option was active the start timestamp in the wav file auxi chunk was not including the initial padding, which screwed up the time sync in SDR Console.


ka9q:
Looks ok, but please continue testing.

PR #88, opened 2025-01-05
Fix pcmrecord diagnostic output time when channels > 1

scottnewell:


ka9q:
Another conflict. I recently removed the SSRC from the logging messages because they're already part of the file name, and I'm trying to tighten frequent log messages.

ka9q:
I think i've got this one. I'm always getting confused about whether I'm counting individual PCM samples, or PCM frames (which for stereo is two samples). I should be more consistent in my variable naming.

PR #89, opened 2025-01-05
Add max_length option to pcmrecord

scottnewell:
Much like the existing lengthlimit option, but max_length does not pad the wav file at startup, and max_length exits when the all wav files are completed.

My use case for this option is something like the FMT. I'd like to be able to create a cron entry to record a receiver channel group for an hour or so. The existing lengthlimit option will continue to create new files, which I don't need. lengthlimit will also pad the file at startup, which I also don't need. If you think this use case is too obscure, please don't hesitate to close without merging.

### DIFF
--- a/pcmrecord.c
+++ b/pcmrecord.c
@@ -136,11 +136,13 @@ struct session {
   int64_t samples_written;
   int64_t total_file_samples;
   int64_t samples_remaining;   // Samples remaining before file is closed; 0 means indefinite
+  bool complete;
 };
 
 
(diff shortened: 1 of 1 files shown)


ka9q:
There's a conflict here, apparently with a performance enhancement I added recently. When linearly searching a linked list, move the found item to the top of the list so another lookup in the near future will find it more quickly. Can you resolve this and resubmit?

scottnewell:
I'll have to resubmit as a new pull request. It shouldn't conflict with the session linked list shuffle.

PR #90, opened 2025-01-05
Fix pcmrecord fp32 wav file fact chunk sample/frame count value

scottnewell:


ka9q:
I'm not using that wav chunk, so I'm glad you found and fixed it.

PR #91, opened 2025-01-06
Add --details arg to powers to add bin/freq/newline to output

scottnewell:


ka9q (approved):
Seems reasonable. Thanks.

PR #92, opened 2025-01-08
Fobos SDR Support + Change the way SDRPlay is compiled

cmh716:
This main purpose of this PR is to add support for the Fobos SDR, which can sample up to 80Ms and has a 40MHz+ spectrum.  This radio requires a driver, libfobos, which must be built from source before compiling because there is not yet a precompiled package.

Originally, I planned to make a Makefile.fobos, but quickly observed some problems with this approach:
1. I would not be able to compile support for both the SDRPlay and Fobos since SDRPlay was built from Makefile.sdrplay
2. As changes occurred to Makefile.linux and Makefile.debug, these same changes would need to be performed by the owner of the SDRPlay code (KE5GDB) and myself, KC2DAC, for Fobos so they would frequently break for no real benefit since 99% of the contents of Makefile.sdrplay/fobos are the same as Makefile.linux.

For example, the Makefile.sdrplay is **currently broken** because of recent changes requiring libsamplerate.

My proposal in this PR request is to remove Makefile.sdrplay and instead incorporate some simple logic into the main Makefile.linux and Makefile.debug to optionally compile SDRPlay and/or Fobos support.  I hope that @ke5gdb is okay with this. 

If we want to compile KA9Q-RADIO with default set of SDRs, we would run: make (as we do now)
If we want to compile KA9Q-RADIO with default set of SDRs + FOBOS, we would run: make FOBOS=1
If we want to compile KA9Q-RADIO with default set of SDRs + SDRPlay, we would run: make SDRPLAY=1
If we want to compile KA9Q-RADIO with default set of SDRs + FOBOS + SDRPlay, we would run: make FOBOS=1 SDRPLAY=1

If we don't set these optional parameters, it would be business as usual, and these would be skipped


cmh716:
Thanks for the feedback @ke5gdb .  I think this will result in less rework for all of us.

cmh716:
I have both a SDRPlay and Fobos and both seem functional when I compiled via:
make FOBOS=1 SDRPLAY=1 

This is without a Makefile.sdrplay

ke5gdb:
Also @argilo and I looked at putting the SDRPlay driver in the CI pipeline but ran into some issues due to, you guessed it, that damn closed source API. At some point I'd like to investigate using [libmirisdr](https://github.com/ericek111/libmirisdr-5), but that's for another day.

cmh716:
yeah, the Fobos API is open source, but no pre-compiled package exists, so it needs to be built from the source. Initially, I had the Makefile configured to automatically fetch and build the source (and it worked), but I figured there would be too much apprehension having that much change to support one more SDR.  I decided to go this (much simpler) route until Rigexpert creates a proper libfobos package.  Once they do, it'll be simple to add Fobos in the mainline.  I'm not sure how much interest there will be in Fobos in KA9Q-RADIO, but for VHF-UHF, it's the highest bandwidth device available for this project.

fventuri:
Couple of notes:
- another reason for writing `Makefile.sdrplay` is that I don't own an RTL-SDR, an Airspy, a FunCube or an RX-888, and I don't want to have to install additional libraries and packages just to run ka9q-radio with my SDRplay RSPdx and RSPduo, so I stripped the original `Makefile.linux` of all the other types of SDRs
- also a while ago (I see it's about 3 years ago) I reorganized the source files for ka9q-radio so that it could be configured and built using `cmake` and therefore be a little easier to select the SDRs ones wants to have built. You can find what I did in the `cmake` branch here: https://github.com/fventuri/ka9q-radio/tree/cmake

Franco

cmh716:
Thank you for the explanation, Franco, and for contributing to this project!

cmh716:
Phil added a config validation yesterday, so I've updated fobos.c and sdrplay.c to stay current

ka9q:
Oh man, I guess I have to address this problem soon. I haven't used cmake much before, but it's becoming clear that ka9q-radio's build options are becoming complex enough to really need it. Yeah, it seems inelegant to have to install a package you don't actually need just to get something to compile, but at the same time wouldn't it be nice to have a binary that just runs everywhere? That may not be important as long as people continue to compile from github sources, but _eventually_ I need to package all this up in Debian packages, and then a more general binary is still desirable. I could just use cmake and turn on all the options when making the Debian package, but then apt would still have to install those extra dependencies.

fventuri:
Phil,
another option could be to use `dlopen()` to load the required 3rd party libraries at runtime as needed: https://tldp.org/HOWTO/Program-Library-HOWTO/dl-libraries.html
With this approach there would be only one general ka9q-radio binary, and it wouldn't be linked to say libsdrplay_api and therefore it wouldn't add that dependency for those who don't have a SDRplay RSP.
It would add some code to dynamically load the symbols for the few functions specific to SDRplay API, but I think there's just a dozen or less of them that I invoke in ka9q-radio `sdrplay.c`, so it is not a big deal. If you are interested, I can change the `sdrplay.c` source code to do that.
This is very similar to what Leif Åsbrink does in Linrad.

Franco

ka9q:
That's an excellent suggestion, Franco. I was pondering dynamic linking but had forgotten about dlopen() as I don't think I've ever used it.

ka9q:
Franco, I just read dlopen() and realized it actually looks (!) easy. I've already done most of the work required in having a set of pointers to functions in my Frontend structure for the setup, start, tune and other functions. The library would then implement those functions with the names name_setup, name_start, etc. The only remaining question is what the shared library file should be called. I could make that a keyword in the [hardware] section of the config file or have it default to some convention.

ka9q:
I just pushed over a main.c with dlopen support for hardware devices. _It's untested_. Franco, give it a look and tell me if it meets your needs.

fventuri:
Phil,
these days I am answering emails and messages using a laptop that has very little installed.
My main PC and the SDRplay SDRs are all in boxes that I haven't had the time to unpack yet (I just brought them in from the pod this morning).
Once I have everything setup, I'll be able to test it and I'll let you know how it goes, but it'll be a few days.

On the other hand, I just thought that it might be easy for you to reorganize the code for the RX-888 as a dynamic library and just try with that, and see how it goes (it would also give me a template about what to do for the SDRplay code).

Franco

ka9q:
OK, I thought of doing just that but figured I'd let you do the honors. If it'll be a while I'll try it when I get a chance. I've been on a long hacking binge lately.

ka9q:
After a few glitches, it works!  Rather than remove the static drivers from radiod and risk breaking existing installations, I've added a testing kludge to the makefile that builds a separate dynamic version of each existing driver (rx888.so, airspy.so, etc). To test them, remove the corresponding references in main.c and rebuild radiod to force the loading of the dynamic version. Add the line "library=/usr/local/lib/airspy.so" in the hardware section of your config file to load it.

This should give the SDRPlay and Fobos fans all the hooks they need to create dynamically loadable drivers. Have at it!

ke5gdb:
Awesome! I'll start testing it with the SDRplay this week.

ka9q:
If I'd known it was going to be this easy, I would have done it long ago. Thanks Franco for the suggestion!

cmh716:
Awesome thanks Phil and Franco!  I'll be testing fobos right away.  One unrelated question, Phil

On your new key validation (for example):
config_validate_section(stdout,Dictionary,"hardware",Airspy_keys,NULL);
  
Shouldn't it be "airspy" instead of "hardware"  because in the airspy conf file, it's [airspy]?

I only ask because I added it to my program and to get the warning messages on bad config entries, I needed to use "fobos" so I want to make sure I did not do anything wrong.

cmh716:
it works!  I suppose if we can to add fobos to this project now, all I need to do is supply:
fobos.c
config/an_example_fobos.conf (with library in the config)
a markdown document of how to install the driver and build/copy the shared object.  

I'll get on that.  this will be easy.  Thanks!

ka9q:
Right! Glad to hear it works. We may have to decide how to structure the makefiles for all this stuff.

ka9q:
Yes indeed, that was a typo. I'd noticed this and intended to fix. Will do so now.

ka9q:
Fixed. I also moved the checks to after verifying that the device= entry is correct, though it doesn't really matter because that's a fatal error anyway.

cmh716:
This is what I've come up with to add Fobos:

https://github.com/cmh716/ka9q-radio/compare/main...cmh716:ka9q-radio:fobos_shared_object

This is just a comparison view.  Do these steps seem appropriate?  It seems to work in my tests.

cmh716:
Or, do you think I should make a tiny Makefile.fobos for this?

ka9q:
Looks reasonable to me. Can you send me a push request here?

cmh716:
Thank you, Phil.  I have just reset this branch so only those changes are included now in this pull request.  I have not included any of the SDRPlay stuff in here but if anyone wants me to do the same with SDRPlay, I'll gladly do it and submit a different request.

ka9q:
OK, I'm not sure how to handle this, still being a git neophyte, but I can see what you did in my Makefile.linux and I'm making those changes by hand. I wanted to just accept your doc files but don't know how.

ka9q:
I'll also accept any new C files you want to add to the repository, since they shouldn't affect anything else if you don't use them.

ka9q:
I just extracted a few and manually added them myself. If you have any more, send them over. It's OK that fobos.h is missing for now, because the makefile doesn't try to compile fobos.c

ka9q:
I got the SDRPlay working! I've had a RSP1 and a RSP2 sitting on my shelf for years, never used because of the driver status. I cleaned up the existing sdrplay.c, made changes to Makefile.linux to build it as a dynamic module. And I found and installed the API -- that was the hard part. I eventually found https://github.com/srcejon/sdrplayapi and installed it. I now have my unit receiving the two local National Weather Service broadcasts using one of the sample config files somebody (Franco?) wrote. The sdrplay.so file is not built by default because of the hassle involved in installing the API package, and the fact that it contains proprietary object-only code in a driver. I have no idea what it does but it seems to work.

To build with sdrplay support, first install the API. Then say "make SDRPLAY=1", then "sudo make SDRPLAY=1 install".

PR #93, opened 2025-01-12
Fix spectrum demod bin level scaling

scottnewell:
Fairly confident on the 1/N scaling factor, less so on the 2. I did test with a function generator at various levels and with this fix powers, baseband, and ka9q-web outputs all matched to better than 1 dB.

### DIFF
--- a/spectrum.c
+++ b/spectrum.c
@@ -84,10 +84,10 @@ int demod_spectrum(void *arg){
       for(int i=0; i < bin_count; i++){ // For each noncoherent integration bin above center freq
 	double p = 0;
 	for(int j=0; j < binsperbin; j++) // Add energy of each fft bin that's part of this user integration bin
-          p += gain * cnrmf(chan->filter.out.fdomain[binp++]);
+          p += cnrmf(chan->filter.out.fdomain[binp++]);
 
 	// Accumulate energy until next poll
-	chan->spectrum.bin_data[i] += p;
+	chan->spectrum.bin_data[i] += (p * gain);
       }
     }
   } while(downconvert(chan) == 0);
--- a/spectrum.c
+++ b/spectrum.c
@@ -79,10 +79,12 @@ int demod_spectrum(void *arg){
       chan->filter.min_IF = -chan->filter.max_IF;
     } else {
       int binp = 0;
+      float gain = (2.0 / (float) N);   // scale each bin value by 2/N (and hope N isn't 0!)
+      gain *= gain;                     // squared because the we're scaling the output of complex norm, not the input bin values
       for(int i=0; i < bin_count; i++){ // For each noncoherent integration bin above center freq
 	double p = 0;
 	for(int j=0; j < binsperbin; j++) // Add energy of each fft bin that's part of this user integration bin
-	  p += cnrmf(chan->filter.out.fdomain[binp++]);
+          p += gain * cnrmf(chan->filter.out.fdomain[binp++]);
 
 	// Accumulate energy until next poll
 	chan->spectrum.bin_data[i] += p;


ka9q:
Sorry I haven't been following this, I've been deep in hack mode with some of my own features. See my estimate_noise() function. In there is a comment to the effect that with a standard FFT, a round trip IFFT(FFT(x)) gives you x scaled by N, which means the power is scaled by N^2. My convention is to allocate half of that to the FFT and half to the IFFT, so the power in the spectrum is scaled by N from the input. I think there also has to be a correction for the overlap in the FFT input, as you are otherwise counting some of the input samples more than once. I even have a conditional piece of code enabled by #define PARSEVAL that uses Parseval's theorem to verify my calculations: the integrals of the energies in the time and frequency domains are the same. I had to do this in my noise estimation because I wanted it to be calibrated; the code in spectrum.c was only intended to give a relative power so I didn't do a correction. I'm really glad you worked on this.

scottnewell:
I was so excited that I didn't notice that obvious optimization. I'll rework the patch shortly.

I did try to enable the Parseval test, but it depends on a deleted Frontend.reference variable, and I wasn't sure how that all worked.

No idea on how to correct the overlap--understanding that is way past my ability.

My current ka9q-web assumed the scaling error was at the input to the forward FFT, and I found the "For efficiency, signal levels now scaled to full A/D range" comment in radio.h confusing--does that still apply?

ka9q:
Thanks for catching that comment about gain scaling. I remember going back and forth on that some time ago. I think that right now the input samples are scaled by the number of A/D bits and the front end analog gain, i.e., full range and 0 dB should give 1 unit RMS in the floating point representation. But I'll have to check that. The dynamic range of even 32-bit floating point is huge by radio standards, so it hardly even matters where you apply overall receiver gain (as long as you do it somewhere).

ka9q:
I'm putting this in. Please continue to check the accuracy of the power calculation.

The overlap problem is actually very simple. When performing fast convolution (the core of ka9q-radio), the input A/D sample buffers fed to the FFT overlap partially in time, e.g., each FFT contains 80% new data and 20% old data. When you do the IFFTs in the individual channels to convert back to the time domain, you discard the first 20% of each output block. So 20% of the samples are being counted twice when you look at the powers in the frequency domain, and I think that means you have to correct your estimates by 10*log10(0.8) = -0.97 dB. Try it.

Gory background details: Filtering is convolution. The filter convolves the input signal by the filter's impulse response. Convolution in the time domain corresponds to simple multiplication in the frequency domain (i.e., multiplying the signal spectrum by the frequency/phase response), so you can efficiently filter with the FFT. But the FFT does circular convolution when you want linear convolution; the tail of the convolution "wraps around" and contaminates the first part of the output. The solution is to overlap the input windows so you can simply discard the wrapped-around, contaminated part of the output. The overlap duration limits the impulse response duration of whatever filter you want to apply, and this is a system-wide configuration. The sharper a filter the longer its impulse response, so each channel is limited in the sharpness of its filters unless it also performs another stage of filtering (which I just added as the experimental "filter2" feature.

It's really appealing to reuse the frequency domain signal as a spectrum analyzer, but there are some more caveats besides the "double counting" problem (which is easily corrected anyway). Normally, a spectrum analyzer applies a "sampling window" in the time domain before performing the FFT; this knocks down the sidelobes at the expense of broadening the main lobe. (You should see these sidelobes in ka9q-web if you zoom way in on a strong narrow carrier.) You do get some benefit from the existing time-domain overlap in that the frequency bins are more closely spaced than their widths, eg, 50 Hz wide bins every 40 Hz instead of 50. (This is another way to see how you're double-counting some of the power). This does help reduce the "scalloping loss" associated with a rectangular (i.e., no) sampling window;you won't lose as much power when a signal falls between two frequency bins.

I've been thinking of how to improve our reuse of the frequency bin data for a spectral display. It's probably fine when zoomed all the way out, but if you want good resolution when zoomed way in it would probably be better to just create a receiver channel for the desired range with IQ output, then perform a more conventional windowed FFT on that to get the data you want. I may put this into spectrum.c.

PR #94, opened 2025-01-12
Fix CI

scottnewell:
Needed to add libsamplerate, and MacOS wouldn't build unless config.c included string.h. This seems to be enough to make the boxes green again in my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,8 @@ jobs:
             libiniparser-dev \
             libavahi-client-dev \
             portaudio19-dev \
-            uuid-dev
+            uuid-dev \
+            libsamplerate0-dev
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Setup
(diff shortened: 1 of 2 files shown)


rrobinett:
I found that on some OSs libsamplerate-dev was not available or apt replaced it with libsamplerate0-dev which does support ka9q-radio but generates warning messages each time you run WD.
It would help WD users if a more widely supported library could be used

scottnewell:
Just to be clear, this PR is only intended to fix the Github continuous integration stuff. I don't think the ci.yml is used anywhere else, is it?

ka9q:
I've fixed the workflows myself. I've always ignored them in the past since I compile locally on both Debian Linux and Macos with port (not brew), and I've been deleting a lot of failed runs from my mailbox. I don't know yml, nor do I run Brew (or some old obscure versions of Ubuntu) so I'm never sure how to fix these things.

A few minutes ago I did discover, quite fortuitously, that these scripts distinguish between tabs and spaces; a file that looks perfectly good to the eyes will still fail with an error message pointing to the wrong line (because it was continued over multiple lines). Years ago I would have said that people who write utilities like that should be shot, but I'm mellowing in my old age.

PR #99, opened 2025-01-25
Workaround for the :59.x files created by pcmrecord in -L -j mode

scottnewell:
I'm seeing pcmrecord with the -j and -L flags create filenames that are not an integer multiple of the -L FileLengthLimit value.

I've found a timespec that causes the issue and included it as a test case with the PCMRECORD_TESTING #define.

For example, with now.tv_sec = 1737742319 and now.tv_nsec = 742568973, the filename would have been 20250124T181159Z_15000000_iq.wav instead of 20250124T181200Z_15000000_iq.wav.

time now = 2025-01-24T18:11:59.7Z
file time = 2025-01-24T18:11:59.7Z
Rounding: convert 1737742319 (18:11:59.7) to 1737742320; test with now.tv_sec = 1737742319; now.tv_nsec = 742568973; creating '20250124T181200Z_15000000_iq.wav' 16000 s/s stereo f32le 15000000.000 Hz iq

### DIFF
--- a/pcmrecord.c
+++ b/pcmrecord.c
@@ -976,6 +976,14 @@ int session_file_init(struct session *sp,struct sockaddr const *sender){
   }
   struct timespec now;
   clock_gettime(CLOCK_REALTIME,&now);
+
+#ifdef PCMRECORD_TESTING
+  #pragma message "WARNING: This will force weird file times, and it will break the output files!"
+  // debug only: force weird rounding filename glitch?
+  now.tv_sec = 1737742319;
+  now.tv_nsec = 742568973;
+#endif
+
   struct timespec file_time = now; // Default to actual time when length limit is not set
   sp->file_time = file_time;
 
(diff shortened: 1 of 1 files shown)


ka9q (commented):
I'm not sure I want to do this. Rob has been obsessed with these file timestamps, and I keep telling him that it only matters that they be monotonically increasing so he can concatenate them in proper order. (I still can't figure out why he insists on making life hard for himself instead of just making all the various sized files he wants directly with parallel pcmrecord commands.) The timestamp on each new file in a segmented recording is meant to be the _actual clock time_ when the new file is started, because this detects drift between the sample clock rate and real time. By changing that you lessen the use of that information.

I do see that I am truncating instead of rounding the seconds to nearest when I use "JT format", and I'm also truncating the 1/10 second field when I use my regular name format, so it's reasonable that I add code to do that. But I don't think I want to do what you're proposing here. Rob just has to be convinced that he can't eat his cake and have it too.

ka9q:
I've pushed over my own changes that round the file creation time to the nearest second for JT-format file names and to the nearest tenth of a second for my own file name format. Hopefully correctly. If Rob is unhappy with this I will stop naming files with timestamps altogether and use sequence numbers. I will not create intentionally incorrect timestamps.

I also fixed my handling of the --mintime option, as it was keeping a lot of short files that should have been deleted. I was counting silence in each segment, and I forgot to reset the segment duration counter when a new file started on the same channel.

PR #100, opened 2025-01-29
Fix pcmrecord memory leak

scottnewell:


ka9q:
I was going to say "good catch!" but then I looked closer. Note the FREE(frontend->description); line in decode_status.c. FREE() is a macro that calls free on its argument, then nails the pointer to NULL. Are you still sure there's a leak?

scottnewell:
Valgrind found it, not me.

pcmrecord allocates a frontend structure on the stack, sets it to zero, and calls decode_radio_status(). Since it's zeroed, nothing for decode_radio_status() to free. But the description still gets allocated by decode_string() and returned to pcmrecord, which then does a shallow copy of the frontend struct containing a pointer to that malloc'd string into each session. And next time through the input loop it happens again. I'm not seeing anywhere the allocated string is ever free'd.

ka9q:
Ok, then it _was_ a good find. Your report prompted me to spend a whole evening with -fanalyze. Lots of complaints about dereferencing NULL pointers, nearly all of which were resolved with an assert(pointer != NULL) immediately after malloc/calloc/etc. There were a bunch of false alarms with FILE pointers, and it couldn't follow the alloc/free logic in some usb routines (not sure I could either). But it never found this one.

I think I intended to make 'description' a static array a while ago, and this prompted me to finally do it. I'm still trying to overcome habits developed 50 (!) years ago when memory was scarce. Also I am never certain just how big to make something, so dynamic allocation let me punt that issue. But yeah, keeping structures shallow-copyable whenever possible is a good idea, and I'm going to try to get rid of any other dynamic pointers when I can.

PR #102, opened 2025-01-31
Fix github CI file by removing illegal tabs

scottnewell:


scottnewell:
Tried and failed to get onion+web building on Macos, too. vsyslog() ended up beating me.

PR #103, opened 2025-02-01
pcmrecord: free resequencing queue allocations at program exit

scottnewell:


ka9q:
I'm not sure there's an "official" rule to free stuff right before program termination since everything will get freed anyway. But if it shuts up valgrind, fine with me.

PR #104, opened 2025-02-02
Switch back to ncursesw to fix UTF-8 encoding/display in control

scottnewell:
I don't know if there are any side effects, or why it was switched from ncursesw to ncurses in commit
fb63136d70ed0f05f65e650d3e1047bed55bc022. This patch does seem to fix the weird character display issue with control.


ka9q:
THANK YOU! This one was driving me nuts.

PR #106, opened 2025-02-07
Link to Hombrew's ncurses in CI

### LINKED ISSUE #105: ncursesw not present on macOS
The macOS CI build began failing with 118e8465dccef0218b4c793a6ef7be3f82463570, which replaced ncurses with ncursesw in Makefile.osx. It appears that ncursesw is not supported on macOS: https://invisible-island.net/ncurses/ncurses.faq.html#what_platforms

argilo:
Fixes #105.

Homebrew's ncurses is a so-called "keg-only" package, which does not symlink into the usual locations to avoid overriding the system-provided ncurses package. Thus, we need to opt into using it by providing the correct paths to the compiler, which I've done here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,6 @@ jobs:
       - name: Setup
         run: ln -s Makefile.osx Makefile
       - name: Compile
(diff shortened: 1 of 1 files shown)


ka9q:
I'm glad you guys are maintaining the CI stuff. Thanks!

PR #107, opened 2025-02-08
Split -march=native into a separate variable

argilo:
ka9q-radio builds with `-march=native`, presumably to maximize performance. This option is only safe to use when the code will be executed on the same machine where it is compiled. Auto-RX ships precompiled binaries which must run on any machine, so it needs to remove the `-march=native` option. At present, it does that by overriding the whole `COPTS` variable:

https://github.com/projecthorus/radiosonde_auto_rx/blob/35d951590db45be213e1a1eba8303324637460ae/Dockerfile#L65-L68

This is not ideal, since the override needs to be kept in sync with ka9q-radio's version. I think it would be useful to split this into its own variable, which I've done here. ka9q-radio can then be built portably with `make ARCHOPTS=`.

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -17,10 +17,10 @@ ifeq ($(DYNAMIC),1)
      DOPTS += -DFORCE_DYNAMIC=1
 endif
 
-
+ARCHOPTS = -march=native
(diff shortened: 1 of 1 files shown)


ka9q:
Thanks, seems reasonable. If you can think of how to automate this further, let me know.

PR #112, opened 2025-02-11
Widen the rate column

argilo:
When receiving an FM stereo channel, the data rate is 1536.0 kb/s, which makes the rate column butt up against the pt column:

Screenshot from 2025-02-10 23-14-29

Increasing the width to 7 solves the problem.

### DIFF
--- a/monitor-display.c
+++ b/monitor-display.c
@@ -684,7 +684,7 @@ static void update_monitor_display(void){
   // Data rate, kb/s
   if(x >= COLS)
     goto done;
-  width = 6;
+  width = 7;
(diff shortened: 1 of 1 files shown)


ka9q:
I guess an alternative would be to get rid of the trailing '.0'...

argilo:
That would be fine too. Feel free to disregard this PR if you'd like to do it another way.

ka9q:
Instead of widening the field I omit the decimal point and tenths if the number is >= 1000.

ka9q:
I wish there was a way to specify the max width of a converted number instead of the minimum.

argilo:
The new version looks good. :+1:

PR #114, opened 2025-02-11
Remove zeroing of pthread_t

argilo:
As far as I can tell, zeroing and zero-checking of `pthread_t` objects was added in an attempt to solve a threading bug (joining non-existent threads, which was later solved in fa1130f5165f0315ccc68b36f2f9f9d35b290d0a).

The zeroing doesn't really solve the bug, and just hides its presence. So I think ka9q-radio would be better off without it.

### DIFF
--- a/airspy.c
+++ b/airspy.c
@@ -291,7 +291,6 @@ int airspy_setup(struct frontend * const frontend,dictionary * const Dictionary,
 }
 int airspy_startup(struct frontend * const frontend){
   struct sdrstate * const sdr = (struct sdrstate *)frontend->context;
-  ASSERT_ZEROED(&sdr->monitor_thread,sizeof sdr->monitor_thread);
 #if 0
   // This should work, but it doesn't
   // So we set it in the first callback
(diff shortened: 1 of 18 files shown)


ka9q:
Yeah, I've been trying to find a portable way to check for uninitialized thread descriptors and there doesn't seem to be one. But these are just asserts; do they really harm anything?

argilo:
ASAN catches joins of uninitialized threads, and in fact any uses of uninitialized values. Valgrind can do the same. That's why I don't see value in littering the code with these checks.

argilo:
Also, I don't think they're the right check really: initializing a pthread_t to zero doesn't make it safe to call pthread_join. Only calling pthread_create does.

argilo:
Note: There were a couple other usages of `ASSERT_ZEROED` in filter.c which I left in, since they were unrelated to threading. I'll leave it to you to decide whether they are worth keeping. If they were intended to catch bugs, perhaps ASAN could do the job.

PR #115, opened 2025-02-11
Improve stereo separation

argilo:
WFM stereo reception suffers from poor stereo separation (only about 14.7 dB). I believe the main reason is that the `lminusr` complex filter selects only one of the two (identical) sidebands of the real-valued composite signal, and therefore only gets half the power. Multiplying the output by sqrt(2) solves the problem, and improves stereo separation to 36.1 dB.

### DIFF
--- a/share/presets.conf
+++ b/share/presets.conf
@@ -70,8 +70,8 @@ conj = no
 # wideband broadcast FM stereo. Output forced to 48 kHz stereo; defaults to North American de-emphasis time constants
 demod = wfm
 samprate = 384k
-low =  -100k
-high = +100k
+low =  -110k
+high = +110k
 threshold-extend = no ; not implemented
 deemph-tc = 75.0    # microseconds; change to 50.0 outside North America & Korea
 deemph-gain = 0.0   # dB, not necessary for WFM because of high corner frequency
--- a/wfm.c
+++ b/wfm.c
@@ -205,7 +205,7 @@ int demod_wfm(void *arg){
       for(int n = 0; n < audio_L; n++){
 	complex float subc_phasor = pilot.output.c[n]; // 19 kHz pilot
 	subc_phasor = (subc_phasor * subc_phasor) / cnrmf(subc_phasor); // square to 38 kHz and normalize
-	float subc_info = __imag__ (conjf(subc_phasor) * lminusr.output.c[n]); // Carrier is in quadrature
+	float subc_info = sqrtf(2) * __imag__ (conjf(subc_phasor) * lminusr.output.c[n]); // Carrier is in quadrature
 	assert(!isnan(subc_info));
 	assert(!isnan(mono.output.r[n]));
 	// demultiplex: 2L = (L+R) + (L-R); 2R = (L+R) - (L-R)
--- a/wfm.c
+++ b/wfm.c
@@ -89,7 +89,7 @@ int demod_wfm(void *arg){
   // FCC says +/- 2 Hz, with +/- 20 Hz protected (73.322)
   struct filter_out pilot;
   create_filter_output(&pilot,&composite,NULL,audio_L, COMPLEX);
-  set_filter(&pilot,-20./Audio_samprate, 20./Audio_samprate, chan->filter.kaiser_beta);
+  set_filter(&pilot,-100./Audio_samprate, 100./Audio_samprate, chan->filter.kaiser_beta);
 
   // Stereo difference (L-R) information on DSBSC carrier at 38 kHz
   // Extends +/- 15 kHz around 38 kHz


argilo:
I added in a couple more improvements:

1. The pilot filter was so narrow that the filter skirts overlapped with 0 Hz and attenuated the pilot significantly. This seems to cause bad performance sometimes. Widening to +/- 100 Hz preserves the pilot signal.
2. Widening the WFM filter from +/- 100 kHz to +/- 110 kHz gives an additional 4 dB improvement in stereo separation (to 40.7 dB). This seems to be the point of diminishing returns; increasing to +/- 120 kHz only added another 0.4 dB.

ka9q:
Oh wow, you've really dug into this. I have hardly used the wfm demodulator since I wrote it. I used to work in broadcasting so I know FM stereo, but I didn't really know the gains that should be used in a DSP implementation to put the signals together. So I stopped when it seemed to work and I didn't refine it further.

Thanks for your improvements!

argilo:
I have worked on the FM stereo implementations in Gqrx and gr-rds in the past, so I thought I would take a look over ka9q-radio's as well. One difference is that Gqrx and gr-rds use a PLL to track the pilot. I haven't investigated what impact that might have on performance.

It's nice to have broadcast FM support built into ka9q-radio, since it is a good "hello world" project when setting up and testing an SDR system.

ka9q:
Yeah, I thought about using a PLL, as I already have one in ka9q-radio, but I figured a narrow filter should work well as long as there isn't too much noise or interference. I made the filter tight to keep any noise down, but I didn't really test its performance. Or the gain of the L-R channel. I didn't quite follow why the sqrt(2) is needed but I agree it sounds much better now.

ka9q:
I made a minor tweak to your fix: I substituted the math constant M_SQRT2 just to make sure the compiler doesn't actually compute the square root. It's cast to float to ensure the multiply isn't done with double precision. (Gnu C has M_SQRT2f but it's not portable).

ka9q:
I don't understand your comment about not using both sidebands of the difference signal. I first spin them down from 38 kHz to baseband and apply a -/+/-15 kHz filter, preserving both sidebands. Then I multiply them by the frequency doubled pilot, normalized to unity amplitude. I was unsure of the proper gain so I just used unity. Obviously this wasn't the right value but I still don't really understand why. Back in the analog days I would just tweak a potentiometer with a screwdriver until it sounded best...

argilo:
By "both sidebands" I meant 38 kHz and -38 kHz from the composite signal.

To be honest I didn't do the math, I just noticed that the gain value which maximized the stereo separation matched sqrt(2) to three significant figures. It seemed like a plausible value. :smile:

Gqrx and gr-rds work with real-valued signals and the correct gain factor there is 2.

argilo:
I tried the math with complex values and it seems like a gain of 2 should be needed there too. I'll have to think about where an extra factor of sqrt(2) might come from.

ka9q:
My filters can provide any combination of real/complex in and real/complex out. In this case, the FM baseband signal is real and I get complex out so I can shift it down to zero and extract the real-only L-R signal. Because the pilot is not coherent with my sampling clock, the baseband DSB L-R signal will slowly rotate, so I slowly rotate it to the Q axis by the doubled pilot frequency (I'm not quite sure why it's Q rather than I). More to the point, I've often run into 3 dB (sqrt 2 amplitude) or 6 dB (2 * amplitude) factors especially when the output of a filter is complex and the input is real, or the reverse. I handled this empirically; see the gain computation in filter.c:set_filter(). I guess I should revisit this. It might be helpful to apply Parseval's theorem here; the total energy in the time domain equals the total energy in the frequency domain)

argilo:
> I'm not quite sure why it's Q rather than I

From https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.450-4-201910-I!!PDF-E.pdf:

> 2.2.2.5 The relative phase of the pilot signal and the sub-carrier is such that, when the transmitter is modulated by a multiplex signal for which A is positive and B = -A, this signal crosses the time axis with a positive slope each time the pilot signal has an instantaneous value of zero. The phase tolerance of the pilot signal should not exceed ±3° from the above state. Moreover, a positive value of the multiplex signal corresponds to a positive frequency deviation of the main carrier.

argilo:
> More to the point, I've often run into 3 dB (sqrt 2 amplitude) or 6 dB (2 * amplitude) factors especially when the output of a filter is complex and the input is real, or the reverse.

Aha! Maybe it's the mono signal that's the problem. It's calculated with a real to real filter, and if I'm reading filter.c correctly a 1/sqrt(2) gain is applied in that case (because the output type is real). Maybe there shouldn't be any scaling in that case?

ka9q:
You may be right. I'm going to test Parseval's theorem for all four combinations. What complicates things is that a real signal is really the sum of mirror image positive and negative frequency spectra, but real signals have only half that power. So should I apply a sqrt(2) factor or not? This is giving me a headache.

ka9q:
The wikipedia article also says that the multiplexed signal can be thought of as alternating between left and right at 38 kHz, which is apparent when you look at this formula shown there (dunno if this will render). I had never thought of it that way, and it suggests another decoding method. It also says that the amplitude ratio is unity.

    [ 0.9 [ A + B 2 + A − B 2 sin ⁡ 4 π f p t ] + 0.1 sin ⁡ 2 π f p t ] × 75   k H z {\displaystyle \left[0.9\left[{\frac {A+B}{2}}+{\frac {A-B}{2}}\sin 4\pi f_{p}t\right]+0.1\sin 2\pi f_{p}t\right]\times 75~\mathrm {kHz} }

argilo:
> It also says that the amplitude ratio is unity.

The (peak) amplitude ratio is unity, but the power ratio is 1/2, because the L-R signal is multiplied by a sine wave. The receiver then loses another 1/2 power by filtering out an image of the stereo signal, which is why a gain of 2 (4x power) is needed to restore the proper amplitude of the L-R signal.

PR #116, opened 2025-02-12
Fix time constant calculation

argilo:
The FM time constant displayed in `control` is incorrect due to an error in the calculation, which I've fixed here.


argilo:
I verified that the correct time constant is displayed for WFM and PM channels.

ka9q:
Hah! I think we need a mutex on finding bugs. I made my fix before I saw this PR. I did it a little differently, using log1p().

PR #117, opened 2025-02-12
Add SANITIZE=1 option to makefile

argilo:
Address Sanitizer (ASAN) and Undefined Behaviour Sanitizer (UBSAN) are useful for identifying bugs in C programs, and they found several of the bugs I have reported to ka9q-radio. So I thought it might be useful to add a `SANITIZE=1` option to the makefile so that others can easily use them.


ka9q:
Good idea. I should use it too.

PR #120, opened 2025-02-12
Fix stereo subcarrier gain

argilo:
As explained in https://github.com/ka9q/ka9q-radio/pull/115#issuecomment-2653519666, a gain of 2 must be applied to the L-R subcarrier after demodulation. With this change in place, stereo separation returns to ~40 dB.

### DIFF
--- a/stereod.c
+++ b/stereod.c
@@ -499,7 +499,7 @@ void *decode(void *arg){
 	  // zero PCM input would cause a divide-by-zero and a NAN result
 	  // that would poison the de-emphasis integrators if we didn't check for it
 	  subc_phasor /= a;
-	  left_minus_right = __imag__ (conjf(subc_phasor) * stereo.output.c[n]); // Carrier is in quadrature with modulation
+	  left_minus_right = 2.0f * __imag__ (conjf(subc_phasor) * stereo.output.c[n]); // Carrier is in quadrature with modulation
 	}
 	  
 	float left = mono.output.r[n] + left_minus_right; // left channel = L+R + L-R
--- a/wfm.c
+++ b/wfm.c
@@ -205,7 +205,7 @@ int demod_wfm(void *arg){
       for(int n = 0; n < audio_L; n++){
 	complex float subc_phasor = pilot.output.c[n]; // 19 kHz pilot
 	subc_phasor = (subc_phasor * subc_phasor) / cnrmf(subc_phasor); // square to 38 kHz and normalize
-	float subc_info = __imag__ (conjf(subc_phasor) * lminusr.output.c[n]); // Carrier is in quadrature
+	float subc_info = 2.0f * __imag__ (conjf(subc_phasor) * lminusr.output.c[n]); // Carrier is in quadrature
 	assert(!isnan(subc_info));
 	assert(!isnan(mono.output.r[n]));
 	// demultiplex: 2L = (L+R) + (L-R); 2R = (L+R) - (L-R)


ka9q:
As per your suggestion I removed the unnecessary gain scaling for real output filters in filter.c:set_filter(), doesn't that solve the problem more cleanly/

argilo:
> As per your suggestion I removed the unnecessary gain scaling for real output filters in filter.c:set_filter(), doesn't that solve the problem more cleanly?

Prior to that change, the L+R (mono) channel had a gain of 1/sqrt(2) applied (in the filter), and the L-R (stereo) channel needed sqrt(2) gain to get it up to the right level. Now the L+R channel has a gain of 1, so the L-R gain needs to be increased as well.

> The formula for the composite multiplex signal given in the wikipedia article shows that the actual gain should be unity.

That formula shows how the transmitted signal is assembled, but if you go through the math for the receive chain you'll see that the recovered L-R signal ends up at half the amplitude.

argilo:
One way to think about it is that the sin(4*pi*fp*t) term is actually the sum of two complex exponentials of amplitude 1/2, and the downconversion process filters out one of the two.

ka9q:
OK, I should get more sleep...

ka9q:
OK, I think I get it now. In the (real only) composite baseband signal the L+R signal occupies both 0 to +15 and 0 to -15 kHz. The L-R signal occupies +/- 15 kHz around both +38 and -38 kHz, i.e, there are actually *four* copies of the difference signal spectrum. By rotating the complex version of the L-R signal down 38 kHz, the copy around -38 kHz gets shifted down to -76 kHz, out of the way. I didn't understand at first what you meant by the "two sidebands" of the difference signal, I thought you meant I was somehow only using 38 to 53 or 23 to 38 kHz. Now I'm wondering about that 90 degree rotation of the reconstructed 38 kHz subcarrier.

argilo:
> Now I'm wondering about that 90 degree rotation of the reconstructed 38 kHz subcarrier.

Note that the formula in the wikipedia uses `sin` for the pilot and L-R signals (and not `cos`). You could replace those with the real parts of complex exponentials, but you'd have to start at -90 degrees. Squaring the pilot would move it to -180 degrees, 90 degrees out of phase from the L-R signal.

ka9q:
Now I'm wondering if I have the channels correct or swapped. I don't have a stereo test generator, as you can probably guess. Maybe I could add one to sig_gen.

argilo:
I'm using this GNU Radio flow graph:

[stereo_tx.grc.txt](https://github.com/user-attachments/files/18774434/stereo_tx.grc.txt)

It transmits two tones on the left channel for five seconds, then the right channel, then both channels, then both channels out of phase. Note: It does not apply pre-emphasis at present.

The sample RDS transmitter from gr-rds should also correctly encode stereo: https://github.com/bastibl/gr-rds/blob/maint-3.10/examples/rds_tx.grc

I used Audacity's "Analyze --> Measure RMS" feature to measure the output level of each channel.

argilo:
One gotcha when using GNU Radio to generate FM stereo: It's important to use a single 19 kHz Signal Source block to generate both the pilot and any required multiples of the pilot frequency (by squaring the pilot). If two or more independent Signal Source blocks are used to generate different frequencies, they will drift out of phase due to floating point error in their phase accumulators.

I imagine it would be handy to have FM stereo built into sig_gen.

argilo:
> Now I'm wondering if I have the channels correct or swapped.

I did test that, and it's correct as far as I can tell.

ka9q:
Thanks for all your help here. This is something I threw together years ago with minimal testing, and haven't used much. But yeah, it's a great "hello world" program for ka9q-radio, to use your analogy. Re accumulated phase errors, yeah, this is something I looked into a long time ago when I wrote osc.c. I minimize them with double precision, but I still have to do occasional amplitude normalization. A good reminder that even doubles aren't perfect. I seem to recall an academic paper showing a different way to generate complex exponentials that isn't as subject to numerical precision but it looked complex (hah) and double precision seemed good enough at the time.

PR #131, opened 2025-02-21
Fix compilation with glibc < 2.34

### LINKED ISSUE #130: ka9q-radio fails to build with glibc < 2.34
libpthread was integrated into glibc in version 2.34 (released August 2021). As a result, it is no longer necessary to explicitly link to it with `-lpthread`.

In e410528ed3ea82d5aad3cf313f06a1d9e4a69936, `-lpthread` was removed, which means that ka9q-radio no longer builds with earlier glibc versions:

```
cc  -o aprs aprs.o libradio.a -lbsd -lm
/usr/bin/ld: libradio.a(misc.o): in function `realtime':
misc.c:(.text+0x249): undefined reference to `pthread_getname_np'
/usr/bin/ld: misc.c:(.text+0x298): undefined reference to `pthread_getname_np'
/usr/bin/ld: libradio.a(misc.o): in function `norealtime':
misc.c:(.text+0x3be): undefined reference to `pthread_getname_np'
/usr/bin/ld: misc.c:(.text+0x430): undefined reference to `pthread_getname_np'
collect2: error: ld returned 1 exit status
make: *** [Makefile.linux:122: aprs] Error 1
```

Thus older OSes like Debian 11 and Ubuntu 20.04 are no longer supported.

Was this change intentional? If not, then `-lpthread` should be added back for programs that need it. It may also be worth extending CI to build ka9q-radio on multiple OS versions to catch accidental breakage of this sort.

argilo:
Fixes #130.

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -119,81 +119,81 @@ dynamics: $(DYNAMIC_DRIVERS)
 # Executables
 
 aprs: aprs.o libradio.a
-	$(CC) $(LDOPTS) -o $@ $^ -lbsd -lm
+	$(CC) $(LDOPTS) -o $@ $^ -lbsd -lm -lpthread
 
 aprsfeed: aprsfeed.o libradio.a
-	$(CC) $(LDOPTS) -o $@ $^ -lbsd -lm
+	$(CC) $(LDOPTS) -o $@ $^ -lbsd -lm -lpthread
 
 control: control.o libradio.a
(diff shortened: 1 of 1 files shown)


ka9q:
I was wondering why it compiled without -lpthread with gcc. I guess it doesn't in some environments. Not every component in ka9q-radio uses threads but I guess it doesn't hurt to add it to all.

PR #144, opened 2025-04-05
Fix CI build for CMake 4

argilo:
CI builds began failing when the GitHub Actions runner updated to CMake 4, which requires `cmake_minimum_required` to be set to 3.5 or higher. The "onion" dependency specifies 2.8:

```
Cloning into 'onion'...
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

Here I've added the suggested override, and the CI run now passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,8 @@ jobs:
           cd onion
           mkdir build
           cd build
-          cmake -DONION_USE_PAM=false -DONION_USE_PNG=false \
+          cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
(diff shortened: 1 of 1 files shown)


ka9q:
Sorry to take so long, I was on the road and lost some inertia here.

PR #146, opened 2025-04-20
Fix spectrum bin level so it doesn't vary with bin width

scottnewell:
The bin power seems to vary based on bin width. I'm testing with an arb set to 2.5 MHz at ranges from -50 to -10 dBm, and I see the baseband and spectrum power match only when bin width is 40 Hz. Wider bins show less power. Removing the scale by bin count factor seems to fix this. Or it's closer--the baseband and powers readings now match.

### DIFF
--- a/spectrum.c
+++ b/spectrum.c
@@ -188,7 +188,7 @@ int demod_spectrum(void *arg){
 	  outf += ratio;
 	  count++;
 	}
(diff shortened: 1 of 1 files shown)


ka9q:
OK, I wasn't sure what to do here. I was normalizing the energy per hertz, but maybe I was doing that twice?

PR #147, opened 2025-04-20
Change spectrum switchover point from >50 to >= 40 Hz

scottnewell:
40 Hz is an allowed bin width in direct mode, right?

### DIFF
--- a/spectrum.c
+++ b/spectrum.c
@@ -77,7 +77,7 @@ int demod_spectrum(void *arg){
     int bin_count = chan->spectrum.bin_count <= 0 ? 64 : chan->spectrum.bin_count;
     float bin_bw = chan->spectrum.bin_bw <= 0 ? 1000 : chan->spectrum.bin_bw;
 
-    if(bin_bw > 50){
+    if(bin_bw >= 40){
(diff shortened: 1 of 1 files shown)


ka9q:
Yes, it is, but I am thinking about windowing. Do you know about sample window theory? Take a look at the FFT section of spectrum.c. When bin width < 50 Hz I perform two 50% overlapping FFTs with a Kaiser window applied to each one, a classic method for spectral analysis (though many different windows exist.) This reduces sidelobes much more quickly than with a rectangular (i.e, no) window, which is the case for the big forward FFT in the fast correlator. At very wide spectral bin widths, the sidelobes from the forward FFT, although high, are all close to the main peak so they don't affect the resulting bin sum much. But they become more significant as the spectrum bin width decreases to the (usually) 40 Hz width of the big forward FFT.

I was thinking that by making an earlier (higher bin width) transition to the channel-to-FFT analyzer as the resolution bandwidth decreases, I could reduce the artifacts associated with the lack of windowing on the big FFT. I am actually thinking about *increasing* the transition frequency from 50 Hz, but I first want to see how much it costs in CPU.

PR #149, opened 2025-04-24
Include "ssrc" key in list of valid channel keys.

scottnewell:
I was getting a "key "ssrc" not found" warning at startup. This seems to suppress it.


ka9q:
Good catch, thanks.

PR #150, opened 2025-04-29
Add BladeRF support.

askulysh:
This patch adds BladeRF support. Radiod is able to run with 61.44MHz sample rate on a modern nettop.  Several cards are supported also.
Tested with BladeRF 2 xA4

Also fixes small bug in squelch closing.


ka9q:
Looks good, though I don't have a blade RF myself. Thanks!

PR #171, opened 2025-07-28
Add RX888 config option to skip USB reset

scottnewell:
I can't get any alternate FX3 firmware to work if the USB reset is present. Franco's rx888_stream program does work with alternate FX3 firmware and it does _not_ do the USB reset. Weird.

I've added a "hack_no_usb_reset" config option with the default to perform the USB reset. Not sure if this is the correct approach...comments?

### DIFF
--- a/rx888.c
+++ b/rx888.c
@@ -106,6 +106,8 @@ struct sdrstate {
   pthread_t agc_thread;
 };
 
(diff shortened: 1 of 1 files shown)


ka9q:
OK, if this breaks anything it's on you... :-)

PR #173, opened 2025-08-04
Add support for HydraSDR RFOne

### LINKED ISSUE #170: Hydrasdr support request
Would it be possible to add HydraSDR support?
It looks a lot like the AirSpy R2.
I'd be happy with it.
https://github.com/hydrasdr

argilo:
Closes #170.

Here I have added support for HydraSDR RFOne, which is optionally compiled if libhydrasdr is installed or `make HYDRASDR=1` is specified.

Because libhydrasdr is nearly identical to libairspy, the new hydrasdr.c is mostly a copy of airspy.c, with the `airspy_` prefix replaced with `hydrasdr_`.

I did a quick test receiving FM radio, and ka9q-radio ran fine with the HydraSDR. I tested firmware versions 1.0.0-rc2 and 1.0.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,15 @@ jobs:
             libliquid-dev \
             avahi-daemon \
             libgnutls28-dev libgcrypt20-dev
+
+            cd /tmp
+            git clone --branch v1.0.2 https://github.com/hydrasdr/rfone_host.git
+            cd rfone_host
+            mkdir build
+            cd build
+            cmake ..
+            make -j4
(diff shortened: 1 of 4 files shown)


ka9q:
I mean, it even advertises itself to USB with the same vendor and product codes, and airspy_info lists it.

ka9q:
There are probably some residual conflicts to resolve here since I restructured Makefile.linux a couple of days ago. In dynamic build mode (the default) everything is now compiled with -fPIC and the separate -pic.o files are gone. It didn't seem worth the complexity.

PR #175, opened 2025-08-17
add direct_sampling option to RTL-SDR

jonsnow1357:


ka9q:
Thanks, I don't use the rtl-sdr much myself so I'm glad you did this.

PR #179, opened 2025-09-06
docs and gitignore updates

jonsnow1357:


jonsnow1357:
love the src folder :+1: 

do you want the doxygen config back?

bvernoux:
Regarding HydraSDR, please note that the correct brand name is "HydraSDR" (or "hydrasdr") written as one word, not "Hydra SDR" or just "hydra" as currently shown in files like docs/SDR/hydra.md.
The first SDR product is specifically called "HydraSDR RFOne" it's important to use the complete product name since HydraSDR is a brand that will encompass multiple products, including the upcoming HydraSDR TGMCK, which isn't an SDR but rather a multiple GNSSDO in-phase clock system (with additional features) designed for multi-SDR, phased array, and laboratory applications.

PR #181, opened 2025-09-14
Add RX888 FX3 firmware images with watchdog support

scottnewell:
20s_watchdog.img: reset to DFU mode if >20 s with no USB transfers
60s_watchdog.img: reset to DFU mode if >60 s with no USB transfers
120s_watchdog.img: reset to DFU mode if >120 s with no USB transfers
120s_failure.img: used to verify watchdog will reset to DFU mode


ka9q:
Thanks, Scott! Sorry it took so long.

Issue #186, opened 2025-11-26
Get stable timing information into the RTP streams/ RTCP sender reports. Also into pcmrecord output.


pjsg:
I'd like to get accurate timing information into the `pcmrecord` output. Probably using the BWF extension (https://tech.ebu.ch/docs/tech/tech3285.pdf#page=9) to the WAV file format. This is just another information block that defines an accurate timestamp of the first sample in the file. It actually includes the number of samples since midnight, along with a standard date/time. The result is still readable as a WAV file.

For `pcmrecord` to be able to do this, it needs to extract the timing information from the RTP stream -- but that is just a sample number since some start time. RTCP sender reports contain the mapping from sample number to absolute time (only I don't think that the current code generates that information correctly). 

The absolute time of the USB transfers from the RX-888 are generated, though they don't flow through the filtering and demod phase. I have no idea what the delay inside the RX-888 is, but I suspect that it is < 50μS. I haven't thought about the other front-ends.

I'd like to be able to get (from the pcmrecord output) the timestamp of the signal to within 1mS. Using the WWV transmissions would allow determination of whether the end-to-end flow works. For me, the light speed delay from WWV is around 12mS -- and I like to see a number around that value.

pjsg:
I did look -- and it looks like it is working fine. Thank you.

PR #187, opened 2025-11-30
Adds support to pcmrecord to add the 'bext' chunk to WAV files if RTCP is enabled

pjsg:
This is a first attempt at solving (part of) the improved timing of capture files (issue #186 ). I think that it works, but the problem is that the radio.c `rtcp_send` doesn't construct the sender report correctly. The fields around this are all badly named -- the rtp_timestamp field in the sender report corresponds to the timestamp field in the RTP data packet -- and this is actually measured in units of samples. 

I beleive that the code change in pcmrecord.c does implement this correctly, but I need to figure out how to adjust the code in radio.c to send the right information. [To be honest, I'm very unclear on how to compute the processing delay from receiving the samples from the frontend all the way to assembling them into a packet]

If there is no RTCP being sent by radio.c, then the bext chunk will not be added (it is actually replaced by a JUNK chunk).

### DIFF
--- a/src/pcmrecord.c
+++ b/src/pcmrecord.c
@@ -1848,12 +1848,17 @@ static int end_wav_stream(struct session *sp){
   header.StartMillis = (int16_t)(((sp->file_time / 1000000) % 10));
 
   if(sp->has_sr && sp->has_rtp_timestamp){
+    uint32_t ntp_msw = sp->first_sr.ntp_timestamp >> 32;
+    uint32_t ntp_lsw = sp->first_sr.ntp_timestamp & 0xffffffff;
+
     snprintf(header.bext.Description, sizeof(header.bext.Description),
-             "RTCP SR: NTP=%lx RTP=%u PC=%u BC=%u CNAME=%.40s NAME=%.40s TOOL=%.40s",
-             (unsigned long)sp->first_sr.ntp_timestamp,
+             "RTCP SR: NTP=%x.%x RTP=%u PC=%u BC=%u RTPT=%u CNAME=%.40s NAME=%.40s TOOL=%.40s",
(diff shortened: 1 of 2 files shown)


pjsg:
This is the code from `radio.c` that constructions the ntp_timestamp and the rtp_timestamp:

```c
    // Construct NTP timestamp (NTP uses UTC, ignores leap seconds)
    {
      struct timespec now;
      clock_gettime(CLOCK_REALTIME,&now);
      sr.ntp_timestamp = ((int64_t)now.tv_sec + NTP_EPOCH) << 32;
      sr.ntp_timestamp += ((int64_t)now.tv_nsec << 32) / BILLION; // NTP timestamps are units of 2^-32 sec
    }
    // The zero is to remind me that I start timestamps at zero, but they could start anywhere
    sr.rtp_timestamp = (0 + gps_time_ns() - Starttime) / BILLION;
```
I think that the last line is missing a multiply by the sample rate -- but that isn't right either as it doesn't take into account the delay through the processing  chain and any inaccuracies caused by the sampling rates not being exactly right. I also don't know if the `0` is actually right either as it ought to depend on when the first rtp data packet is sent.

I got really confused when I started to think about leap seconds and the fact that the occasional day has more samples than other days.

pjsg:
I was hoping that a reasonable estimate of the delay could be acquired from monitoring WWV (after compensating for lightspeed delays). I agree that adding a GPS sync pulse would be better, but it requires more circuitry!

ka9q:
Have you talked with Michael Hauan, AC0G and Scott Newell N5TNL? I know Scott has already done some cool stuff with WWV seconds ticks and minute beeps.

PR #197, opened 2026-02-18
Cast to a 'float' was causing a dropping of right channel audio when a stereo signal was present for wfm

Deatojef:
For WFM, it looks like right channel audio was getting dropped because of the cast to a ‘float’ instead of ‘float complex’.


Deatojef:
Took a lot of searching to find this and ultimately cried uncle - had Claude Code look at wfm.c.  Initial tests seemed to be positive when listening to area FM broadcast stations as audio was coming from both channels after fix.

PR #203, opened 2026-04-09
wd-record v0.16: add freq-path option, fail on /dev/shm full

scottnewell:
With the --freq-path option in -W mode, the wav files will be placed in subdirectories named for the channel frequency.

Try and detect if there's no shared memory available and fail, rather than throwing an unexpected bus error at startup.

Rename diagnostic shared memory segment from pcmrecord.bpsk.pid to wd-record.pid.

Bump version to 0.16.

### DIFF
--- a/src/wd-record.c
+++ b/src/wd-record.c
@@ -329,6 +329,7 @@ static bool filter_500 = false;
 static bool leaky_folding = false;
 static double leaky_folding_filter = 0.99;
 static bool notch_filter_initialized = false;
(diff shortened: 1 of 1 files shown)


ka9q:
It bombs on macos because it doesn't support posix_fallocate(). I'll install a portable stub.

ka9q:
do you still want me to work on this PR? Or can you submit a new one based on the latest version of wd-record.c?

PR #206, opened 2026-04-24
spectrum: poll-driven narrowband mode when client requests fft_avg=1

on8st:
## Summary

For narrowband spectrum channels (`rbw <= crossover`), `demod_spectrum()` calls `downconvert()` once per block (~50 Hz). Inside `downconvert()`, `execute_filter_output()` runs an inverse FFT to materialise the baseband window for the analysis FFT. With many always-on narrowband channels this IFFT dominates radiod's CPU. On a Pi 5 driving 21 always-on bands across ~6 MHz total span at 5 Hz poll rate (a typical multi-band waterfall workload), `radiod` jumps from ~110 % to ~388 % CPU — almost all of it spent in those per-block IFFTs whose output is then thrown away by client-side averaging.

The per-block IFFT only earns its keep when (a) someone is going to look at every frame, or (b) the analysis runs over many frames (`fft_avg > 1`) and so needs a continuous baseband stream. When the client opts into `SPECTRUM_AVG=1`, the analysis FFT consumes exactly one window of baseband per response — which we can produce on demand at poll time.

This PR adds a "lazy" path to `demod_spectrum()`:

- When `fft_avg <= 1` and no poll is pending, skip `downconvert()` entirely and sleep one block period.
- When a poll arrives, run a small burst of `downconvert()` calls (capped at 4 blocks / ~80 ms) to refill the ring buffer with fresh baseband, then run `narrowband_poll()` as usual.

Channels with `fft_avg > 1` (the compiled-in default of 10, or any client that explicitly opts into radiod-side averaging) keep the existing continuous-IFFT behaviour — the `lazy_eligible` guard requires `fft_avg <= 1`, so this path is strictly opt-in via `SPECTRUM_AVG=1`.

## Numbers

Pi 5 / RX888 / 21 narrowband channels @ 5 Hz polls, client also patched to push `SPECTRUM_AVG=1`:

| setup                              | radiod CPU | delta vs idle |
| ---------------------------------- | ---------- | ------------- |
| radiod alone (no clients)          | ~110 %     | baseline      |
| radiod + 21 channels (legacy)      | ~388 %     | +278 %        |
| radiod + 21 channels (this patch)  | ~129 %     | **+19 %**     |

~93 % of the per-channel overhead removed, with no observable change in spectrum frame rate or content (frames continue to arrive at the requested poll rate, ages stay sub-100 ms).

## Implementation notes

- `Blocktime` is in seconds (`Frontend.L / Frontend.samprate`), not milliseconds. The lazy sleep computes `tv_nsec` from a seconds basis (`* 1e9`) — a microseconds basis here would degenerate into a busy-wait and erase the savings.
- The 4-block burst cap bounds added latency at ~80 ms while still letting `narrowband_poll()` see a window of fresh baseband even when `fft_n > sampcount`.
- No protocol or default behaviour change: existing clients (which leave `fft_avg` at its default of 10) take the original code path.

## Test plan

- [x] Build clean on Pi 5 (Debian).
- [x] Verify CPU drop with 21-band narrowband workload at 5 Hz polls.
- [x] Verify frame rate / age / contents match legacy path with `SPECTRUM_AVG=1`.
- [x] Verify a co-resident `ka9q-web` client (which uses `fft_avg=10`) still works and keeps the legacy continuous-IFFT path.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/spectrum.c
+++ b/src/spectrum.c
@@ -92,18 +92,32 @@ int demod_spectrum(void *arg){
 
     // fairly major reinitialization required
     if(chan->spectrum.fft_n <= 0){
-      if(chan->spectrum.plan == NULL)
-	fftwf_destroy_plan(chan->spectrum.plan); // will be regenerated on first poll
-      chan->spectrum.plan = NULL;
+      // NOTE: setup_complex_fft / setup_real_fft (called via setup_narrowband / setup_wideband)
+      // already destroy any existing chan->spectrum.plan before allocating a new one.
+      // The previous code here did "if(plan == NULL) fftwf_destroy_plan(plan); plan = NULL;"
+      // which (a) was an inverted NULL check (no-op) and (b) NULL-orphaned the live plan
+      // pointer so the inner setup_*_fft would not destroy it -> leaked plan + buffers.
       FREE(chan->spectrum.window); // force regeneration on first poll
       if(chan->spectrum.rbw > chan->spectrum.crossover)
 	setup_wideband(chan);
       else
 	setup_narrowband(chan);
+      // Cache current parameter values so we do not refire reinit every block.
+      // Without this, the local shadow vars (rbw, bin_count, ...) stay at their
+      // sentinel -1 forever, the comparisons above always set fft_n=-1, and we
+      // re-enter this block on every iteration -> per-block FFT plan + filter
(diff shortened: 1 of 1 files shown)


on8st:
Closing this PR — refiling as #207 as an issue with bug description + suggested patch, per maintainer preference. Happy to reopen as a PR if you'd like the change merged that way.

PR #213, opened 2026-05-29
rx888: fix wedge-detection bug and add in-driver FX3 recovery escalation

on8st:
This addresses a mid-stream wedge mode where the FX3 silently stops delivering samples while remaining responsive to USB control transfers. Before this patch, `radiod` sat in the wedge indefinitely; only a physical replug cleared it. Observed twice on a Raspberry Pi 5 / RX888mk2 station in a single day.

## 1. Wedge-detection fix in `rx_callback`

`rx_callback()` updated `sdr->last_callback_time` unconditionally, before the transfer status check. When the FX3 wedges and starts handing back failed/empty transfers, the callback still fires for each one, `last_callback_time` keeps refreshing, and the *"No rx888 data for 5 seconds"* watchdog in `proc_rx888()` can never trip. `frontend->samples` meanwhile stops incrementing, so `agc_rx888()` correctly sees a measured sample rate of 0 Hz and logs it every minute — but has no authority to abort.

Fix: only update `last_callback_time` on a successful, non-empty transfer. The existing `LIBUSB_TRANSFER_NO_DEVICE` → `usb_device_gone` → immediate-exit path is preserved.

## 2. In-driver tiered recovery via FX3 vendor commands

The SDDC FX3 firmware already exposes vendor commands that can recover the chip without external intervention (see `ExtIO_sddc/SDDC_FX3/USBhandler.c`):

| Code | Name | Action in firmware |
|---|---|---|
| `0xAA` | `STARTFX3` | Reset DMA channel, re-enable GPIF SW input |
| `0xAB` | `STOPFX3`  | Disable GPIF SW input, reset DMA channel, flush USB EP |
| `0xB1` | `RESETFX3` | `CyU3PDeviceReset(CyFalse)` — full chip soft reset + USB re-enumeration |

This patch wires them into `proc_rx888()`'s wedge handler as a two-tier escalation:

- **Tier-1a**: `STOPFX3` → 500 ms → `STARTFX3`. In-stream, no USB re-enumeration. Resets the DMA producer-side state — sufficient for stalls caused by transient back-pressure on the `CY_U3P_DMA_TYPE_AUTO_MANY_TO_ONE` channel. ~0.5 s, 30 s grace window for data to resume.
- **Tier-1b**: `RESETFX3` then `exit(EX_NOINPUT)`. Full chip reset; the USB re-enumeration through bootloader (`04b4:00f3`) + fxload makes the libusb handle invalid anyway, so we exit and let the supervisor (systemd / docker / shell loop) restart us. ~1 s of USB time.

`rx_callback()` resets the tier counter to 0 the moment real data flows again, so a one-off recovery doesn't permanently consume the budget.

## Validation

- Compiles clean with `-Wall -Wextra`.
- Both vendor commands validated by manual USB control transfer on a healthy RX888mk2:
  - `STOPFX3 + STARTFX3`: silent DMA reset, no USB-layer event, sub-second.
  - `RESETFX3`: full chip reset; observed `04b4:00f1` → `04b4:00f3` (Cypress bootloader) → fxload → `04b4:00f1` re-enumeration in ~1 s. Functionally equivalent to a physical replug.
- Patched build (against a Docker-packaged downstream fork that shares this `rx888.c`) running in production on a Pi 5 / RX888mk2 station for >18 h at time of submission. Baseline before the patch was ~2 organic wedges per day on the same hardware.

The escalation has not yet been observed firing on a real wedge in the patched binary, because the patched binary has not wedged in its first 18 h. So we know it doesn't regress normal operation; we don't yet know it cures the wedge case it was designed for. The mechanism is plausible from the FX3 data-path architecture (`StartStopApplication.c` configures a `CY_U3P_DMA_TYPE_AUTO_MANY_TO_ONE` channel with no flow control or error callback, and `Pib_error_cb` is commented out), and the worst case is no worse than today's behaviour (clean exit on watchdog).

### DIFF
--- a/src/rx888.c
+++ b/src/rx888.c
@@ -95,7 +95,12 @@ struct sdrstate {
   unsigned int transfer_index; // Write index into the transfer_size array (unused)
   struct libusb_transfer **transfers; // List of transfer structures.
   unsigned char **databuffers;        // List of data buffers.
-  long long last_callback_time;
+  long long last_callback_time;       // ns timestamp of last *successful* bulk transfer carrying data
+
+  // Tiered mid-stream wedge recovery — see commit message for analysis.
+  // 0 = idle, 1 = Tier-1a fired (STOPFX3+STARTFX3), 2 = Tier-1b fired (RESETFX3)
+  int recovery_tier;
+  long long recovery_grace_until;     // ns; while in tier-1a, suppress further watchdog action until this time
 
   // USB transfer
   int xfers_in_progress;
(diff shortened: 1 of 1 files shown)


ringof:
Check/try:

https://github.com/ringof/rx888-firmware

A lot of the wedging/recovery got sorted in this fork of the firmware.

Docs:
https://ringof.github.io/rx888-firmware/

David Goncalves

ringof:
Sorry, release FW here:
https://github.com/ringof/rx888-firmware/releases/tag/v0.1.0

You should be able to drop in and try it.


David Goncalves

ka9q:
Thanks for the detailed report. I will reorder the software watchdog reset so it'll detect a stalled transfer, but do I still need to do the other recovery stuff? I've preferred to just bomb out and let systemd restart us, especially if restarts aren't very common. David's firmware has been running for several days on a bunch of machines I manage and has been pretty solid.

ka9q:
Another reason I hesitate to recover within radiod instead of just exiting and letting systemd recover is that some of my users rely on exact sample timing, ie, they count samples to mark off periods of time in files for analysis. When radiod restarts, the application sees a new RTP stream and knows to reset its mapping of sample count to wall clock time. If I restart the front end inside radiod, the application will see a stall of indeterminate length and may not even notice it.

ka9q:
I'm going to close this now. Dave's firmware seems a lot more stable than the original, and as I explained I prefer to have radiod simply exit on a front end error and let systemd restart it, so that users see new streams and know that the sample count was interrupted.

PR #218, opened 2026-06-15
Adding CMake build system

techn0mad:
(Supersedes PR #188)

This set of changes eliminates the need for the current Makefiles. It provides a working configuration of the CMake build system (which generates Makefiles behind the curtain, based on what the target system is).

Builds for Ubuntu and macOS happen in the GitHub Actions "CI" workflow, and I have tested the FreeBSD build manually, since nobody seems to offer a virtualized, on-demand environment for doing the same sort of automated testing for it.

As I currently don't have any access to any SDR gear, I don't have a good way to verify the builds work, but I have been wondering if it might be possible to use Phil's "siggen" virtual back-end to write some automated "smoke" tests going forward.


ka9q:
Thanks for all this work. I figured I'd have to start dealing with cmake eventually, and you've helped me climb the learning curve. The changes are extensive and there are some conflicts I need to resolve, so it may take a while to appear in the tree.

ka9q:
I've done a somewhat deeper scan through your changes. Implementing them will be a lot of work, and I'm not comfortable with some of them. Eg, I extensively reworked monitor some months ago to use atomic variables to provide lock-free synchronization of output streams. Things became a lot more stable. Are atomics simply not supported on some other platforms? I know they're not as important on the x86 as some other architectures because of how memory references are ordered but I'd still like to maximize generality. I see the writing on the wall, the x86 is probably on its way out, most likely to be replaced by the ARM, about which I know much less.

I see a lot of changes to header includes that I suppose are necessary to make cmake work with multiple targets. I want my code to run on multiple platforms but I just wish there was a way that didn't require so many ifdefs and levels of indirection everywhere.  I probably spend much more time than I should trying to rewrite stuff to be more portable so it won't require quite so many ifdefs and cmake options.

I see a lot of variable type changes, eg, in monitor, especially from double to float. Many of these variables used to be floats but ChatGPT convinced me that double precision is usually just as fast and you get the benefit of reduced intermediate roundoff errors even if the final result is a float. So my general practice now is to use doubles for random scalar math and reserve floats for arrays, especially when they're likely to be processed by SIMD instructions. I'd like your views on this.

techn0mad:
> I've done a somewhat deeper scan through your changes. Implementing them will be a lot of work, and I'm not comfortable with some of them. Eg, I extensively reworked monitor some months ago to use atomic variables to provide lock-free synchronization of output streams. Things became a lot more stable. Are atomics simply not supported on some other platforms? I know they're not as important on the x86 as some other architectures because of how memory references are ordered but I'd still like to maximize generality. I see the writing on the wall, the x86 is probably on its way out, most likely to be replaced by the ARM, about which I know much less.
> 
> I see a lot of changes to header includes that I suppose are necessary to make cmake work with multiple targets. I want my code to run on multiple platforms but I just wish there was a way that didn't require so many ifdefs and levels of indirection everywhere. I probably spend much more time than I should trying to rewrite stuff to be more portable so it won't require quite so many ifdefs and cmake options.
> 
> I see a lot of variable type changes, eg, in monitor, especially from double to float. Many of these variables used to be floats but ChatGPT convinced me that double precision is usually just as fast and you get the benefit of reduced intermediate roundoff errors even if the final result is a float. So my general practice now is to use doubles for random scalar math and reserve floats for arrays, especially when they're likely to be processed by SIMD instructions. I'd like your views on this.

Hi Phil: I hear what you are saying, and it's been a while since I looked at this branch. As a first take, I would say that requiring changes to the existing code other than include/ifdef/etc. directives is probably a bug and is to be avoided if at all possible. I will revisit it and see if I can simplify it (a lot, if possible, and remove any code changes).

I also took a brief look at the atomic variables issue and I can see that GCC on the BSDs supports them for x86 (and presumably also for ARM, but I could not confirm this). Have you seen this library: https://github.com/bdwgc/libatomic_ops ? It may or may not be of use.

ka9q:
I looked at that package. Near the top of the readme it says

IN NEW CODE, PLEASE USE C11 OR C++14 STANDARD ATOMICS INSTEAD OF THE CORE LIBRARY IN THIS PACKAGE.

...which is what I've already done.

lgadallah-korio:
@ka9q I don't disagree with closing this PR, and I have another one that does not use CMake in the works, if you are interested (in the 'portable-init' branch of github.com/techn0mad/ka9q-radio). It also runs a GitHub Actions workflow that builds ka9q-radio in each of Ubuntu, macOS, and FreeBSD (and I was thinking of adding a Linux distro like Alpine that doesn't have systemd, just for completeness).

PR #219, opened 2026-06-28
Servo SDRplay lna-state / RF gain reduction based on overload flag

ke5gdb:
The SDRplay will assert an overload bit if the tuner detects a front-end overload. This pull request adds an option to increase RF gain reduction when the overload is detected. It is intended to function with fast-attack and slow-decay. If the parameter is omitted from the config file, no automatic adjustments of RF gain reduction will be made. 

Context/use case: for our high-altitude balloon launches, the RTL-SDR tuner AGC far outperforms the SDRplay while payloads are being assembled and tested at the launch site. Only after launch and after some distance is established between the SDRplay receiver and the payloads does the overload stop and valid data is received. This pull request fixes that problem and was demonstrated at Field Day to work quite well.

### DIFF
--- a/config/examples/radiod@sdrplay-generic.conf
+++ b/config/examples/radiod@sdrplay-generic.conf
@@ -13,6 +13,7 @@ antenna = Antenna C
 #samprate = 2000000        ; default is 2M
 lna-state = 2
 if-agc = yes
+overload-gr-interval = 30
 
 frequency = 162480000
 
--- a/src/sdrplay.c
+++ b/src/sdrplay.c
@@ -49,6 +49,15 @@ struct sdrstate {
   double scale;
   enum sdrplay_status device_status;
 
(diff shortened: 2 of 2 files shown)


ka9q:
The overload flag is a nifty feature, I didn't know it existed. I usually detect A/D overload by simply looking for maximal samples (== 32767 or < -32767). I only have a couple of old SDRPlays and I rarely use them because of the proprietary driver, so thank you for contributing this. I'll take your word for it that it works...

ke5gdb:
Thanks Phil! It has been working well, mainly in the context of our amateur balloon launches. When we're ground testing just before a flight, the RS41s and Wenet payloads may be only a few feet from the receiver. The receiver is usually in my vehicle or on a portable satellite rig. Shortly after launch, this function reduces gain reduction, and allows the receiver to work well through the whole flight. I've demonstrated it on at least 3 flights now. 

73!

Issue #191, opened 2025-12-15
radiod CPU utilization increased significantly


argilo:
It appears that commit 323ec100a76d0a2c502c7574370b44ce0cbb75c1 has significantly increased the CPU utilization of radiod, from about 140% to 215% on my Raspberry Pi 4 running radiosonde_auto_rx. I presume the extra load relates to radiosonde_auto_rx's use of a spectrum channel. (It periodically runs `powers radiosonde.local -f 403025000.0 -w 800.0 -b 7438 -i 20 -s 40302503 -c 2`.)

argilo:
I heard from @darksidelemm that tuning `point_budget` may help, so I'll experiment with that today.

argilo:
Adjusting `point_budget` definitely makes a big difference. Here's what I measured on my Raspberry Pi 4:

Baseline `radiod` CPU utilization, before executing `powers`: ~74%
After starting `powers`, commit 6571ae1503717abb7016a8e6710fa937f1a60ecd (prior to the changes): ~81%
After starting `powers`, commit f7c227bac88d51ba6a65070441324c6f4f34fa57, `point_budget = 512 * 1024`: ~154%
After starting `powers`, commit f7c227bac88d51ba6a65070441324c6f4f34fa57, `point_budget = 256 * 1024`: ~108%
After starting `powers`, commit f7c227bac88d51ba6a65070441324c6f4f34fa57, `point_budget = 128 * 1024`: ~93%
After starting `powers`, commit f7c227bac88d51ba6a65070441324c6f4f34fa57, `point_budget = 64 * 1024`: ~84%
After starting `powers`, commit f7c227bac88d51ba6a65070441324c6f4f34fa57, `point_budget = 32 * 1024`: ~82%

ka9q:
Hmm, maybe I should make the averaging parameter (the number of consecutive blocks FFTed and averaged) a parameter in each poll instead of my too-clever budget idea.

argilo:
> maybe I should make the averaging parameter (the number of consecutive blocks FFTed and averaged) a parameter in each poll

That sounds like it could be useful for radiosonde_auto_rx, which wants to average over a long period (20 seconds, by default) to detect intermittent transmitters.

ka9q:
[...]

PR #220, opened 2026-06-30
Add logic to flip spectrum on high-side LO receivers

ke5gdb:
Addresses remaining issues in #191


ke5gdb:
Closing because this is redundant with the proper fix.

PR #224, opened 2026-07-21
Audit fixes: security, correctness, and build hygiene

n4hy:
Nine self-contained commits from an audit pass on the current tree.
Each commit builds cleanly (0 warnings) and can be reviewed or dropped
independently. Grouped roughly by risk from lowest to highest:

## Hygiene

- `gitignore: add build products and installed binaries` — quiet `git status` after a `make -j`.
- `remove signal(SIGKILL, closedown) no-ops` — SIGKILL is uncatchable per POSIX; the three signal() calls in pcmrecord/pcmspawn/wd-record were no-ops that suggested cleanup that never runs.
- `rx888: remove dead #if 0 VHF/UHF tuner code` — `rx888_set_tuner_frequency` was marked "Pretty sure this is broken" and every reference was already inside `#if 0`. −78 net lines.

## Correctness

- `fix multiple correctness/safety bugs from audit`
  - `src/Makefile:186` — `setcap` gets `$(DESTDIR)$(bindir)` so Debian package builds stage the capability instead of writing to the build machine's live binary.
  - `src/status.c:decode_string` — return NULL on malloc failure instead of NULL-dereferencing. The three strlcpy callers in decode_status.c and radio_status.c now guard.
  - `src/decode_status.c` TLV parser — three copies of the type/length/value loop hardened: bounds-check the length byte read, reject a length-of-length exceeding sizeof(unsigned) or overrunning the packet, and use subtraction form for value bounds to avoid pointer-arithmetic wraparound.
  - `src/filter.c:modulo` — the helper `x > m ? x - m : x` returns `m` (out-of-bounds by one) when `x == m`. Replaced with `((x % m) + m) % m` which is correct for any int.
  - `src/rx888.c` USB init error path — removed a double-free: `free_transfer_buffers()` already frees the sdr->transfers and sdr->databuffers arrays; the caller was then FREE()ing them again. Matched the working pattern used in rx888_close.
- `harden two edge cases in AGC and audio queue`
  - `src/linear.c` — guard `samples_per_slice` against rounding to 0 (would spin the peak-finder loop) with an unusually large Blocktime.
  - `src/audio.c` — use a temporary around the queue realloc so an OOM failure doesn't leak the old queue.
- `enable -Wint-conversion and fix the three real bugs it catches`
  - `bladerf.c` — `pthread_join(&sdr->main_thread, ...)` was passing `pthread_t *` where `pthread_t` was expected (on glibc, `pthread_t` is an unsigned long, so this was passing an address as a thread ID). Two sites fixed to pass the value.
  - `radio.c` — `process_section` returns `void *` but one error path had `return -1;`. Inert (return value is discarded at the join sites) but a real type violation. Changed to `return NULL;`.

## Security

- `harden pcmrecord --exec substitution and bound multicast rDNS cache`
  - `pcmrecord.c:1167,1202` — the `--exec` `$d` macro substituted `sp->frontend.description` (a network-controlled string received over the multicast status channel) directly into the command handed to `popen()`. A rogue radiod on the LAN could set the description to shell metacharacters and land arbitrary code execution on any host running `pcmrecord --exec ... $d ...`. Every substituted value is now single-quoted with the standard `'\''` escaping.
  - `multicast.c:formatsock` — the reverse-DNS cache was a linked list with head insertion and no eviction; a long-running radiod exposed to many unique source addresses leaked memory forever. Also, two threads racing on the same unresolved address could each allocate and insert their own entry, duplicating it in the list. Converted to a bounded LRU (4096 entries) with a tail pointer for O(1) eviction, and a re-scan after re-acquiring the lock to close the duplicate-insert race.

## Build, packaging, plugin hygiene

- `build: dedupe driver sources; harden radiod@ service; add bladerf udev`
  - `src/Makefile` — base CFILES now lists every .c file including optional drivers (bladerf, fobos, hackrf, hydrasdr, sdrplay, rx888_boot, gauss, fft-gen, ...) so `-include $(DEPS)` picks up header-change rebuild dependencies uniformly. Removed the duplicate `CFILES += foo.c` from each ENABLE_* block.
  - `service/radiod@.service.in` — conservative sandboxing: `NoNewPrivileges`, `PrivateTmp`, `ProtectSystem=strict`, `ProtectHome`, `ProtectKernelTunables`, `ProtectKernelModules`, `ProtectControlGroups`, `LockPersonality`, `RestrictRealtime`, `RestrictSUIDSGID`, `RestrictNamespaces`, `RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX AF_NETLINK`. Skipped `MemoryDenyWriteExecute` (FFTW SIMD dispatch) and `SystemCallFilter` (subtle breakage risk).
  - `rules/54-bladerf.rules` — plugdev/uaccess for Nuand BladeRF USB (original, 2.0 micro, and older pre-Nuand vendor ID). Analogous to the airspy and rx888 rules.
- `radio.c: explicitly NULL optional plugin fn pointers on dlsym miss` — defense-in-depth: on dlerror() for `_shutdown`, `_tune`, `_gain`, `_atten`, explicitly NULL the corresponding pointer. In practice dlsym returns NULL on function-symbol miss so this is belt-and-suspenders, but the code now makes it obvious that callers must NULL-check.

## Verification

- `make clean && make -j` completes with zero warnings on Ubuntu 24.04 with the full ENABLE_* set.
- Each commit tested to build individually.
- Three DSP claims from the audit were investigated and dismissed as false positives (osc.c renorm divide-by-zero — guarded by `is_phasor_init` requiring |x| ≥ √0.9; rx888 abs() on int16 — auto-promotes to int, INT16_MIN fits in int32; rx888 dc_offset — actually a proper leaky integrator α·N ≈ 0.1). No code change for those.
- Have not yet done a live-signal test against the RX888 rig here; happy to run one and report back if useful.

Please push back on any of these you'd rather leave alone; they're independent commits and easy to drop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,39 @@ old
 Temporary Items
 work
 stash
+
+# build products
+*.o
+*.a
+*.so
+*.d
+src/config_paths.h
+
+# built binaries (src/Makefile DAEMONS + EXECS + old/obsolete targets)
+src/aprs
+src/aprsfeed
+src/control
+src/ctcss
+src/cwd
+src/fft-gen
+src/jt-decoded
+src/metadump
+src/monitor
+src/opusd
+src/opussend
+src/packetd
+src/pcmcat
+src/pcmrecord
+src/pcmsend
+src/pcmspawn
+src/powers
+src/radiod
+src/rdsd
+src/rx888_boot
+src/setfilt
+src/set_xcvr
+src/show-pkt
+src/show-sig
+src/stereod
+src/tune
+src/wd-record
--- a/rules/54-bladerf.rules
+++ b/rules/54-bladerf.rules
@@ -1,0 +1,15 @@
+# Nuand BladeRF USB permissions
+# Grants plugdev group access via uaccess so radiod (member of
+# plugdev) can open the device without root. Analogous to
+# 51-airspy.rules and 71-rx888.rules. For host-specific autostart
+# (SYSTEMD_WANTS to a radiod@<instance>.service), add a
+# 55-bladerf-ka9q.rules that matches on the serial number.
+
+# BladeRF (original)
+ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="2cf0", ATTR{idProduct}=="5246", MODE="0660", GROUP="plugdev", TAG+="uaccess"
+
+# BladeRF 2.0 micro
+ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="2cf0", ATTR{idProduct}=="5250", MODE="0660", GROUP="plugdev", TAG+="uaccess"
+
+# Older Nuand (pre-Nuand vendor ID assignment)
+ACTION=="add", SUBSYSTEM=="usb", ATTR{idVendor}=="1d50", ATTR{idProduct}=="6066", MODE="0660", GROUP="plugdev", TAG+="uaccess"
--- a/rules/Makefile
+++ b/rules/Makefile
@@ -4,7 +4,7 @@
 udevdir	      ?= /etc/udev/rules.d
 
 UNAME_S := $(shell uname -s)
-RULES = 20-rtlsdr.rules 51-airspy.rules 51-hydrasdr.rules 52-airspyhf.rules 53-airspy-ka9q.rules 66-hackrf.rules 66-sdrplay.rules \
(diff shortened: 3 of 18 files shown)


ka9q (commented):
These AI code reviews are a mixed bag. Some definite improvements, some gratuitous or overly fussy changes, and some would have actually broken code. I've worked through most of them, thanks again.

ka9q:
I think I've handled all the ones that make sense, so I'm going to close this. Please send another PR if you find more stuff like this to fix, or if I missed something important.

PR #228, opened 2026-07-25
Portable service discovery: add a `dns_sd.h` backend for mDNS

techn0mad:
## Portable service discovery: add a `dns_sd.h` backend for mDNS

Makes ka9q-radio's mDNS-based service discovery build and run on macOS and
FreeBSD, without changing its behavior on Linux. Reference documentation for the
design lives in `docs/PORTABLE-MDNS.md`; this description covers the change set
and its verification status.

### Motivation

On Linux, publish/browse are implemented by `avahi.c` / `avahi_browse.c`, which
`fork`/`exec` the `avahi-publish-service`, `avahi-publish-address`, and
`avahi-browse` CLI tools. Those tools don't exist on macOS or FreeBSD, so
publish/browse fail there — the concrete reason the macOS build was incomplete.
Both platforms ship the `dns_sd.h` (Bonjour / mDNSResponder) API, which speaks
the same mDNS wire protocol, so a second backend covers them with one body of
portable code.

### Approach

A provider-neutral interface (`avahi.h`) backed by either of two build-selected
backends. The Linux `avahi` backend is left byte-for-byte intact; a new
`dns_sd` backend (`mdns.c`) is added alongside and selected automatically per
platform. No consumer of the interface changes.

### Commits

| Commit | Change |
|--------|--------|
| `avahi.h` de-couple | Strip vestigial `avahi-client/*` / `avahi-common/*` includes (no code references any libavahi symbol); make the header self-contained (`#include <stdbool.h>`, fixed include guard). Zero consumer edits. |
| add `mdns.c` | `dns_sd.h` backend implementing `avahi_start` / `avahi_browse` / `avahi_free_service_table`: `DNSServiceRegister` + `DNSServiceRegisterRecord` publish, async→sync `Browse→Resolve→GetAddrInfo` bridge, persistent-registration thread, same `service_tab` single-buffer alloc/free contract and sort/de-dupe policy as the avahi backend. |
| Makefile backend selection | `MDNS_BACKEND` auto-selection (Linux→avahi, else→dns_sd; overridable); backends referenced via `$(AVAHI_PUBLISH_OBJ)` / `$(AVAHI_BROWSE_OBJ)`; `-ldns_sd` linked off-Darwin. |
| docs | `docs/PORTABLE-MDNS.md` reference documentation. |

### Diff surface

Upstream files touched are minimal: `src/avahi.h` (~10 lines) and `src/Makefile`
(backend selection). `src/avahi.c` and `src/avahi_browse.c` are unchanged. New
files: `src/mdns.c`, `docs/PORTABLE-MDNS.md`.

### Verification

- **Publish→browse round trip** smoke-tested on macOS against the live
  mDNSResponder: a service registered via `avahi_start()` (service + A record +
  TXT) is found by `avahi_browse()` with correct name, host, resolved numeric
  address, port, interface, and TXT records — the exact fields `control.c` feeds
  into `getaddrinfo(..., AI_NUMERICHOST)`.
- **No-match path** returns 0 within the time budget; freeing an all-zero table
  is safe.
- Clean under the runtime **AddressSanitizer / UndefinedBehaviorSanitizer**
  bug-detectors on both paths (validates the single-buffer pack/free contract).
- `mdns.o` builds warning-free under project `CFLAGS` (`-Wall -Wextra`);
  `make -n` confirms correct object/flag selection for Darwin, Linux, and
  forced-`dns_sd` configurations.

### Still open/ToDo

- **End-to-end parity test against a running `radiod`** (A/B the `avahi` vs
  `dns_sd` backends discovering a live instance via `control`). Runnable on
  Linux today; a full on-macOS run is gated on the remaining non-mDNS
  portability work (`<bsd/string.h>` compat shim and the `multicast.c`
  Linux-header dependency).

### DIFF
--- a/docs/PORTABLE-MDNS.md
+++ b/docs/PORTABLE-MDNS.md
@@ -111,7 +111,9 @@ DNSServiceBrowse ─► DNSServiceResolve ─► DNSServiceGetAddrInfo ─► ro
 It services the connection socket with `select()` under a bounded time budget
 (exiting early once results stop arriving), collects completed rows into the
 caller's `struct service_tab`, and applies the same sort + de-dupe-by-name
-policy as the `avahi` backend. Each row's strings are packed into a single
+policy as the `avahi` backend, with one addition: where duplicates of an
+instance differ by interface, a non-loopback one is kept (see section 5). Each
+row's strings are packed into a single
 `->buffer` allocation, so `avahi_free_service_table()` frees a row with one
 `free()` regardless of backend.
 
@@ -157,6 +159,16 @@ No binary links both, so the substitution is clean.
   publish mode (writing `/etc/avahi/hosts` and `/etc/avahi/services`) is specific
   to Avahi on Linux; the `dns_sd` backend always registers services live. The
   global remains defined so consumers link unchanged.
+
+- **Browse results prefer a non-loopback interface.** mDNSResponder reports a
+  locally published service on every interface it is visible on, so browsing for
+  a `radiod` running on the same host yields the same instance more than once —
+  typically once on `lo0` and once on the real interface. `control` passes the
+  surviving row's `->interface` straight to `listen_mcast()`/`join_group()`, so a
+  loopback row there would silently isolate it from the network. The de-dupe
+  therefore keeps a non-loopback row when the duplicates differ. This is rarely
+  visible on the `avahi` backend, where `avahi-daemon` does not normally browse
+  loopback.
 
 - **Name resolution is backend-independent.** Role 1 uses `getaddrinfo()`, so
   the `.local` name behavior documented in `docs/NETWORK-NOTES.md` is the same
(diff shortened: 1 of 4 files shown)


ka9q:
Cool, looks interesting. My primary computer is macos so I would also like to have mdns discovery working. I have plans to add it to some of the other utilities too, like monitor. Looks like you're still actively working on this?

techn0mad:
> Cool, looks interesting. My primary computer is macos so I would also like to have mdns discovery working. I have plans to add it to some of the other utilities too, like monitor. Looks like you're still actively working on this?

Hi Phil: What it needs is a bit more "real world" testing. I did what I could with unit testing on macOS and so on, but what would really confirm it would be to run a version with the original Avahi mDNS, and then run a version with this new MDNSResponder-compliant mDNS and verify that the results are the same. AFAIK, there's no additional code required, just testing.

This reminds me: Can you tell me how the siggen driver can be used to run ka9q-radio in a "hardware-less" mode? It could make testing much easier.

ka9q:
Yes, sig_gen is a front end emulator that requires no hardware. As a cute hack, it uses my 'wwvsim' program to put a simulated WWV at 5 MHz. In config/examples/radiod@siggen.conf is an example of how to use it. Note the hardwired path name to wwvsim; I have to figure out how to have my debian package scripts change this.

PR #229, opened 2026-07-25
spectrum.c complex update & SDRplay IF AGC type fix

ke5gdb:
This PR fixes a bug where only the upper half of the spectrum was being returned from powers when a device with a complex front-end. 

Additionally the type-casting of the IF AGC parameters on the SDRplay were removed since they can exceed 255.

### DIFF
--- a/src/sdrplay.c
+++ b/src/sdrplay.c
@@ -1101,9 +1101,9 @@ static int set_if_gain(struct sdrstate *sdr,int const if_att,int const if_gr,int
   } else if(if_agc_rate == 0){  // use AGC scheme
     sdr->rx_channel_params->ctrlParams.agc.enable = sdrplay_api_AGC_CTRL_EN;
     sdr->rx_channel_params->ctrlParams.agc.setPoint_dBfs = if_agc_setPoint_dBfs;
-    sdr->rx_channel_params->ctrlParams.agc.attack_ms = (uint8_t)if_agc_attack_ms;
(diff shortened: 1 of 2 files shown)


ka9q:
Thank you for doing this! I'll pull it as soon as I've had a chance to fully digest it. Working out those frequency bin operations was quite an ordeal (see filter.c:execute_filter_output).

Looking back at it I realized I should have adopted the convention that frequency bins are always in monotonically increasing order. That would have avoided a lot of ad-hoc special cases and tests for crossing DC and wrapping around while decimating. I didn't notice the bug here (I _did_ say it was untested) because I don't currently use any SDRs that produce complex sample streams; I use the Airspy R2 and Hydra SDR for VHF/UHF (which has a raw real output mode) and a rx888 MKii on HF, which uses real direct sampling that already delivers a right-side-up spectrum. If I were to flip the sign of every other complex sample, that would swap the negative and positive frequencies and put DC in the center. If I were to flip every other real sample from the Airspy, this would flip the spectrum right side up (and also shift the apparent LO down by half the sample rate.) I may still do this some day just for the pleasure of ripping out all those special wraparound tests.

ka9q:
I've accepted your patches to sdrplay.c. Working on the spectrum fixes.

ka9q:
I've also added your fix to the complex wideband spectrum bug that wasn't giving negative frequencies. I also added a fix of my own that's based on the loop template I use in the downconverter output filter, but yours is active. Give both a try if you like. This has been hard to test because ka9q-web currently supports only direct sampling front ends like the rx888. Thanks again!

PR #234, opened 2026-08-01
Remove the unused `<uuid/uuid.h>` and the `uuid-dev` dependency behind it

techn0mad:
`src/radio.c` includes `<uuid/uuid.h>`. Nothing in the tree uses it: there are no `uuid_*` calls anywhere, no `uuid_t` or any other uuid type, and no target links `-luuid`. It is the only uuid reference in the source.

The cost is not the include itself but what it drags along. `uuid-dev` is in `debian/control`'s Build-Depends and in the `apt install` line in `docs/INSTALL.md`, so everyone who builds from source installs a package the build has never used.

**Two commits, deliberately separate**

1. Remove the include — one deleted line in `src/radio.c`. Self-contained, and correct whether or not you want the second.
2. Drop the dependency — `uuid-dev` out of `debian/control` and `docs/INSTALL.md`. This is only safe because of the first.

Split so the second can be dropped or deferred without losing the first.

**Testing**

Built and installed on Ubuntu 24.04 with `uuid-dev` not present on the machine — apt had to fetch it previously, so it is not part of the base image and its absence is real. `radiod`, `control` and `monitor` were each run afterwards to print their version banner. Had anything still needed the header, the build would have failed at the `#include`.

One honest gap: this exercises the make build, not the Debian packaging. I have not run `dpkg-buildpackage`, so the debian/control change rests on the source audit above rather than on a package build.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -5,7 +5,7 @@ Maintainer: Phil Karn <karn@ka9q.net>
 Rules-Requires-Root: no
 Build-Depends: make, gcc, debhelper-compat (= 13),
  libbsd-dev, libfftw3-dev, libopus-dev, libiniparser-dev,
- libavahi-client-dev, uuid-dev, libncurses-dev, libusb-1.0-0-dev,
+ libavahi-client-dev, libncurses-dev, libusb-1.0-0-dev,
  portaudio19-dev, libsamplerate-dev, libogg-dev, libairspy-dev,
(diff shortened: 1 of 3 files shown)


ka9q:
By the way, ka9q-radio-hfdl depends on dumphfdl, which depends on uuid-dev. ka9q-radio-all depends on ka9q-radio-hfdl, so it too will bring in uuid-dev.

lgadallah-korio:
Is that a run-time dependency or a link-time dependency?

PR #241, opened 2026-08-08
Security audit

sm7eca:
I made an AI supported audit. The code is compiled and verified on Mac and installed and tested on Raspberry Pi 5 Trixie.
Details in SECURITY_AUDIT_COMPLETE.md
/Arne

### DIFF
--- a/Pi5/configure-ka9q-direwolf.sh
+++ b/Pi5/configure-ka9q-direwolf.sh
@@ -1,0 +1,231 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Configure Raspberry Pi for KA9Q-radio + Dire Wolf APRS receive pipeline.
+# Tested conceptually against the setup established in this session:
+#   RTL-SDR -> radiod -> aprs-pcm.local RTP/PCM -> pcmrecord -> Dire Wolf 1.7
+#
+# Safe behavior:
+# - Installs required packages.
+# - Enables Avahi.
+# - Backs up existing config files before modifying them.
+# - Preserves existing squelch values; only adds defaults if missing.
+# - Creates a Dire Wolf config using stdin, 12 kHz, 2 PCM channels.
+# - Uses AGW TCP 8002 because port 8000 is already used by ka9q-beacon-mon.
+# - Creates a helper command to start the APRS decoder pipeline.
+# - Does NOT enable a systemd service automatically.
+#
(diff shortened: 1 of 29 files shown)


ka9q:
Thanks for the patches, I'm adding some of them. Most seem to be checks on allocation failures, though calling them "critical security vulnerabilities" is a little overblown. Storage allocation in modern 64-bit virtual memory systems rarely fails; how can it? If it ever does, it's usually a sign of something very serious (corrupted heap, etc) that's hard to recover from meaningfully. That's why I only guarded most of them with assertions that get turned off in production builds. I suppose a case can be made that an alloc failure in production code should at least log a message even if it doesn't attempt recovery so the problem can be identified and investigated, so I'm doing that.

ka9q:
Over the past few days I've substantially restructured how radiod initializes itself and how it starts, initializes and stops individual channels, and I think that addresses some of the problems addressed by one of your patches. Because these are major changes I'm not making them directly to my main branch; they're currently in "setup-restructure". If it seems stable after I run it here for a few days I'll merge them back into main. I tend to use branches sparingly and I don't like to let them sit unmerged too long because they become increasingly impossible to merge back, and they can also confuse users over which is the "right" branch to use.

sm7eca:
Thanks for the thoughtful feedback and for taking the patches seriously. I really appreciate your perspective and the reasoning behind your decisions — it’s clear there’s a great deal of experience behind the way you approach the codebase.
I’m glad some of the suggestions were useful, and I look forward to following the continued development of radiod. Thanks again for the excellent work.

ka9q:
I've manually applied some but not all of the patches. Your AI is rather pedantic, especially on "security"; reasonable minds can differ on exactly how one should handle malloc failures in the 64-bit virtual memory age but calling it a "critical security vulnerability" is a bit extreme. My philosophy is that input should be checked for validity in production code, but code logic errors can be tested for with asserts that are only enabled in debug builds, and a malloc failure is arguably a code logic error since it's hard to see how one can happen unless you've corrupted memory, gone into a loop, or done something else really drastically wrong.

It suggested using memcpy rather than unions to do type punning, which I've also read elsewhere. It looks less elegant to me but since it's supposed to now be the recommended way I converted over.

Some of the other patches seemed useful, like makefile detection of Brew. I use only Macports but I don't mind accommodating those who prefer Brew.

PR #245, opened 2026-08-15
Serialize FFTW plan destruction for thread safety

vk4tmz:
## Summary

Serialize FFTW plan destruction so concurrent demodulator teardown cannot
call FFTW plan-destruction routines at the same time.

## Background

While exercising multiple simultaneous receivers, I observed `radiod`
occasionally crash/core-dump during receiver teardown/reconfiguration,
particularly when multiple demodulators were being torn down concurrently.

Investigation highlighted a potential concurrency issue around FFTW plan
destruction.

FFTW documents its planner-related operations as not thread-safe by default.
Plan destruction is part of that interface, so concurrent calls to
`fftwf_destroy_plan()` should be serialized.

I cannot conclusively attribute every observed `radiod` crash to this race,
but the existing teardown path permits concurrent FFTW plan destruction and
therefore appears to violate FFTW's threading requirements.

FFT execution itself remains thread-safe and is not serialized by this change.

## Change

Add a mutex around FFTW plan destruction.

The lock is limited to destruction of FFTW plans; normal FFT execution and
signal-processing paths are unaffected.

## Scope

- No receiver configuration changes.
- No DSP algorithm changes.
- No serialization of FFT execution.
- Only FFTW plan destruction is serialized.

## Validation

Tested with multiple simultaneous KA9Q receivers, including concurrent receiver
creation and teardown.

Since deploying the serialization change in the production receiver environment,
I have not observed any further unexpected `radiod` crashes/core dumps.

Normal receiver operation and decoding have remained unchanged during testing.

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -160,6 +160,19 @@ fftwf_plan plan_c2r(int N, float complex *in, float *out){
   }
   return plan;
 }
+// FFTW only guarantees thread-safe execution of existing plans. Plan creation
+// and destruction must be serialized by the application. Use the same mutex for
+// both sides of the plan lifecycle so one demod cannot destroy FFTW planner state
+// while another demod is creating or destroying a plan.
+void destroy_plan(fftwf_plan *plan){
+  if(plan == NULL || *plan == NULL)
+    return;
+  pthread_mutex_lock(&FFTW_planning_mutex);
+  fftwf_destroy_plan(*plan);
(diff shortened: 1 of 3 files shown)


ka9q:
Oh good catch! I knew that plan creation had to be serialized but it hadn't occurred to me that destroying them should also be serialized. I kinda assumed it was just freeing a block of memory, but being opaque it might have some embedded pointers.

PR #246, opened 2026-08-15
Make pcmrecord stdout survive session rollover

vk4tmz:
## Summary

Allow `pcmrecord --stdout` / cat mode to survive an RTP session rollover when
the same SSRC reappears from a different network source (for example, after
`radiod` restarts and begins transmitting that SSRC from a new UDP source
port).

## Background

Long-running decoder pipelines commonly use `pcmrecord --stdout` to consume a
specific SSRC.

When `radiod` is restarted or a receiver is recreated, the same logical SSRC
can reappear from a different UDP source port.

`pcmrecord` historically associates the stream with the original RTP session.
As a result, the replacement stream can be treated as a new session even
though it carries the same requested SSRC. The original stdout session can
then remain idle and eventually time out, causing `pcmrecord` to exit while
the requested SSRC is already active again.

This was observed while testing recovery of long-running decoder pipelines
across `radiod` restarts.

## Change

For stdout/cat mode, treat the explicitly requested SSRC as the logical stream
identity.

When that SSRC reappears from a new sender tuple:

- rebind the existing stdout session to the new sender
- clear stale RTP/resequencing state
- continue writing to the existing stdout stream

File-recording modes retain the existing tuple-based session behaviour.

## Scope

- Applies only to stdout/cat operation for an explicitly selected SSRC.
- Does not change normal file-recording session handling.
- Does not change RTP packet processing during an unchanged session.
- Keeps the existing stdout consumer/process alive across sender/session
  rollover.

## Validation

Tested with a long-running `pcmrecord --stdout` consumer while recreating the
upstream receiver with the same SSRC.

The replacement stream used a new sender/session tuple and `pcmrecord`
successfully rebound to it while keeping the existing stdout process alive.

Normal stdout streaming continued after the rollover without requiring the
downstream consumer to be restarted.

### DIFF
--- a/docs/utils/pcmrecord.md
+++ b/docs/utils/pcmrecord.md
@@ -16,6 +16,10 @@ This program reads one or more RTP streams from a multicast group and either wri
 
 > write one stream to stdout, default false. If --ssrc is not specified, selects the first one found and ignores the rest.
 
+**--reacquire**
+
+> with --stdout, continue following the selected SSRC if its RTP sender/session changes. If --ssrc was not specified, the SSRC first selected by --stdout is used.
+
 **-d, --directory**
 
 > directory root in which to write files.
--- a/src/pcmrecord.c
+++ b/src/pcmrecord.c
@@ -511,6 +511,30 @@ static void process_status(int fd){
     sp->prev = NULL;
     Sessions = sp;
   }
+  if(sp == NULL && Catmode){
+    // stdout mode follows one logical SSRC, not one sender socket tuple.
+    // radiod may restart and resume the same restored SSRC from a new UDP
+    // source port. Rebind the existing stdout session so its stale pre-restart
+    // tuple cannot later hit the idle timeout and terminate an active stream.
+    for(sp = Sessions;sp != NULL;sp=sp->next){
+      if(sp->ssrc == chan.output.rtp.ssrc)
+        break;
+    }
+    if(sp != NULL){
+      if(Verbose)
+        fprintf(stderr,"ssrc %u sender changed %s -> %s; continuing stdout session\n",
+                sp->ssrc,formatsock(&sp->sender,false),formatsock(&sender,false));
+      // Discard any packets/resequencing state belonging to the old RTP
+      // incarnation. stdout itself remains open and attached to this session.
+      for(int i=0;i < RESEQ;i++){
+        FREE(sp->reseq[i].data);
+        sp->reseq[i].size = 0;
+        sp->reseq[i].inuse = false;
+      }
+      memset(&sp->rtp_state,0,sizeof(sp->rtp_state));
+      memset(&sp->last_active,0,sizeof(sp->last_active));
+    }
+  }
   if(sp == NULL){
     // Create session and initialize
     sp = calloc(1,sizeof(*sp));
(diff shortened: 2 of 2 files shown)


ka9q:
I could do this, but is it a good idea? It's possible to have more than one instance of radiod on a system, each with its own SSRC space. If they have the same IP source address, the only way to tell them apart is by the UDP source port.

Have you looked at the --exec option to pcmrecord? It can launch a new instance of a decoder for each new stream to a given multicast destination group. I use it to launch HFDL, for example. When I restart radiod, new decoders are launched and the existing HFDL decoders eventually see end of stream and exit.

vk4tmz:
Hiya Phil,

Yes, I see the concern about multiple radiod instances sharing a source IP.

Would multiple radiod instances normally be expected to use different multicast destinations? If so, could the destination multicast group + SSRC provide the persistent logical stream identity, while source IP/port identifies a particular RTP incarnation?

The specific case I'm trying to handle is recovery from a radiod crash followed by an automatic restart. The same configured SSRC comes back, but radiod may use a different UDP source port after the restart.

So what I'm really after is restart/reacquisition semantics for an explicitly selected SSRC, rather than globally ignoring the UDP source port.

I haven't looked closely at --exec yet, but my understanding is that it launches a new decoder for the newly appearing stream. In this case I'd ideally like an existing pcmrecord -S <ssrc> consumer to survive/reacquire the stream after radiod restarts.

ka9q:
You can certainly use different multicast destination addresses (many sites do) but they tend to be a limited resource in Ethernet switches with IGMP snooping so I prefer not to preclude the option to use a common target for, eg, all FT8 traffic or all WSPR traffic. Take a look at the --exec option; I think it might actually be an easier way to do what you want to do, particularly if you have a lot of channels of the same type that all need their own instances of a particular kind of recorder or decoder.

ka9q:
We could make this an option that would apply only to specific pcmrecord commands using the --stdout option. I wouldn't want to make it the default behavior without thinking about it more.

ka9q:
Looks good. I can't argue against it if it's an option. But still take a look at the --exec option and see if it can do what you want. I do use --stdout myself, but usually only for testing and ad-hoc stuff. Turnkey production stuff uses --exec.

vk4tmz:
Thanks Phil. I've updated the PR along the lines you suggested.

The session reacquisition behaviour is now explicitly opt-in via a new long-only `--reacquire` option, and is restricted to `--stdout` mode. Without `--reacquire`, the existing `pcmrecord --stdout` behaviour is unchanged.

With `--reacquire`, `pcmrecord` continues following the selected SSRC on its current input if the RTP sender/session changes, keeping stdout open for an existing downstream decoder. `--reacquire` without `--stdout` is rejected.

I also tested both behaviours against an actual `radiod` restart:

- `--stdout --reacquire`: the same `pcmrecord` process/PID remained running across the restart and resumed output when the stream returned.
- `--stdout` without `--reacquire`: `pcmrecord` retained the existing behaviour and terminated when the original session disappeared.

This seems to give us the recovery behaviour we need without changing the default semantics or affecting the `--exec` use case.

Thanks for the suggestion, happy to adjust the option name or semantics if you'd prefer a different approach.

PR #247, opened 2026-08-15
Initialize linear RTP timestamp before first status response

vk4tmz:
### Summary

Initialize the linear receiver RTP timestamp before sending the first status response.

A newly created linear channel could send its initial status response before `chan->output.rtp.timestamp` had been initialized from the current demodulator job position.

This meant the first status packet could advertise an uninitialized/stale `RTP_TIMESNAP`, while the first RTP data packet already used the newly initialized timestamp.

Consumers such as `pcmrecord` use the status clock/RTP snapshot pair to map RTP timestamps to wall-clock time. If `pcmrecord` received the premature status response, it could calculate an incorrect recording start time.

### Observed failure

This was reproducible when creating several dynamic receivers simultaneously.

For example, a recording started at approximately:

`2026-08-16 07:35:22 AEST`

but one `pcmrecord` filename was timestamped approximately:

`2026-08-15T23:50:34.5Z`

which is roughly 2h15m into the future.

The error closely matched:

`initial RTP timestamp / 12000 Hz`

i.e. approximately the current `radiod` uptime, consistent with `pcmrecord` combining a stale/zero initial `RTP_TIMESNAP` with the first initialized RTP timestamp.

### Change

Defer the initial status response until after the first successful downconversion has initialized the RTP timestamp.

The pending response is then sent immediately, preserving normal status behaviour while ensuring the first advertised RTP clock snapshot is valid.

### Validation

Tested with three simultaneously created 12 kHz USB receivers and separate `pcmrecord` instances.

Before the change, different receivers would intermittently receive incorrect filename timestamps around two hours into the future.

After the change, a test started at:

`2026-08-16 08:16:41 AEST`

produced:

- `9998400k2026-08-15T22:16:41.8Z.wav`
- `9998401k2026-08-15T22:16:41.8Z.wav`
- `9998402k2026-08-15T22:16:41.9Z.wav`

All three timestamps matched the actual recording start time within a fraction of a second.

### DIFF
--- a/src/linear.c
+++ b/src/linear.c
@@ -58,8 +58,15 @@ int demod_linear(void *arg){
   pthread_mutex_unlock(&chan->status.lock);
   realtime(chan->prio);
   do {
-    response(chan,response_needed);
-    response_needed = false;
+    // The initial status response must not advertise an RTP_TIMESNAP before
+    // the RTP timestamp has been tied to the frontend block/job timeline.
+    // pcmrecord uses GPS_TIME + RTP_TIMESNAP to timestamp new recordings; an
+    // early response with the zero-initialized RTP timestamp can therefore
+    // produce a bogus filename time when the first RTP packet arrives.
+    if(first_run || frontend->L == 0){
+      response(chan,response_needed);
+      response_needed = false;
+    }
     pthread_mutex_lock(&chan->status.lock);
     // Look on the command queue and grab just one atomically
     for(int i=0;i < CQLEN; i++){
@@ -92,6 +99,11 @@ int demod_linear(void *arg){
       if(Verbose > 0)
 	fprintf(stderr,"%s starting at FFT jobum %u, preset RTP TS to %u\n",chan->name,first_block,chan->output.rtp.timestamp);
       first_run = true;
+
+      // Now that RTP_TIMESNAP is valid, send the initial response that was
+      // deliberately deferred at the top of the loop.
+      response(chan,response_needed);
+      response_needed = false;
     }
 
     // First pass over sample block.


ka9q:
I have gone back and forth on whether to do the response at the top or the bottom of the loop. Recently I redid the way I initialize and start sessions; I was concerned about possible critical sections, but also about ensuring that a status response gets sent quickly to keep 'tune' from retransmitting unnecessarily. That's why I generate it at the top of the loop.

So one possibility is to not delay the first response, but to just not include the RTP_TIMESTAMP field until after the first execution of the downconverter. This would give a channel-creation command like 'tune' a quick acknowledgement but not send a timestamp that is simply wrong. Would this work for you?

ka9q:
I came up with a different fix that I think is cleaner. Give it a try and let me know if it works for you.

vk4tmz:
Thanks Phil — tested your new fix and it works well here.

I first tested it with `ka9q-recorder` creating two simultaneous dynamic linear receivers. Both resulting WAV RTP timestamps matched the actual wall-clock creation time closely.

I then ran a more abusive test: 10 consecutive record cycles, each destroying and recreating both receivers (20 fresh receiver creations total). All 10 runs completed successfully, and all 20 WAV timestamps remained correct and closely aligned between the two receivers, with no recurrence of the large timestamp offset from the original issue.

I also saw no related `pcmrecord` errors or warnings during the test.

So this resolves the issue for my use case. Your fix is cleaner than my original approach, and I'm happy for PR #247 to be closed as superseded.

Thanks again for looking into it!

ka9q:
Great! Glad it worked for you.

PR #251, opened 2026-08-26
Don't strlcpy dest_string onto itself in loadpreset()

### LINKED ISSUE #250: radiod dies with SIGTRAP in loadpreset(): strlcpy() copies dest_string onto itself
## Summary

`loadpreset()` in `src/modes.c` copies `chan->output.dest_string` onto itself with `strlcpy()` whenever a config section has no `data` key. That is an overlapping copy, which is undefined behaviour. glibc tolerates it, so it is invisible on Linux, but macOS's fortified `strlcpy` detects the overlap and raises `SIGTRAP` — `radiod` dies before starting its first channel.

Present on current `main` (`a4b4076d`) at `src/modes.c:493-494`, introduced in 9c2e3ce9 ("major rewrite of initialization and session creation").

## Cause

```c
char const *data = config_getstring(table,sname,"data",chan->output.dest_string);
strlcpy(chan->output.dest_string,data,sizeof chan->output.dest_string);
```

`config_getstring()` ends in `iniparser_getstring(d,buf,def)`, which returns `def` **verbatim** when the key is absent. The default passed here is `chan->output.dest_string`, so in that case `data == chan->output.dest_string` and the `strlcpy()` has `src == dst`.

C11 7.24.2.4 (and the BSD `strlcpy` contract) require source and destination not to overlap, so this is UB on every platform — Linux just doesn't notice.

## Symptom

```
$ radiod -v radiod@siggen.conf
...
Block time 20.000 ms, block samples L=96,000, overlap 5 (20.0%) M-1=24,000 samples, forward FFT size N=120,000 real
FFTW version: fftw-3.3.10-neon
fft: setpriority failed, Permission denied (13)
$ echo $?
133
```

Exit 133 = 128+5 = `SIGTRAP`. From the macOS crash report:

```
exception: {"type": "EXC_BREAKPOINT", "signal": "SIGTRAP"}

Thread 0 (com.apple.main-thread):
  libsystem_c.dylib   __chk_fail_overlap
  libsystem_c.dylib   __chk_overlap
  libsystem_c.dylib   __strlcpy_chk
  radiod              loadpreset
  radiod              loadconfig
  radiod              main
```

Any config section that omits `data` triggers it, so in practice it fires on the first channel section.

## Minimal reproducer

```c
#include <string.h>
#include <stdio.h>
int main(void){
  char buf[64] = "test-pcm.local";
  char const *src = buf;              // what config_getstring() returns as the default
  printf("before\n"); fflush(stdout);
  strlcpy(buf,src,sizeof buf);
  printf("survived\n");
  return 0;
}
```

```
$ cc -O3 -o overlap overlap.c && ./overlap; echo "exit=$?"
before
exit=133
```

## Suggested fix

Skip the copy when it would be a self-copy:

```c
char const *data = config_getstring(table,sname,"data",chan->output.dest_string);
if(data != chan->output.dest_string)
  strlcpy(chan->output.dest_string,data,sizeof chan->output.dest_string);
```

PR to follow. Happy to adjust if you'd rather fix it at the `config_getstring()` level instead — this same "pass a field as its own default" pattern could bite elsewhere.

## Environment

macOS 26.5.2, arm64 (Apple silicon), Apple clang 21.0.0, Homebrew deps, `radiod` built from `main` with the `sig_gen` front end.

mijahauan:
Fixes #250.

`config_getstring()` returns the default pointer verbatim when the key is absent, so a section with no `data` entry makes `data == chan->output.dest_string` and the `strlcpy()` copies the buffer onto itself.

`strlcpy()` is undefined on overlapping objects. glibc tolerates it, so this is invisible on Linux, but macOS's fortified `strlcpy` traps on the overlap and `radiod` dies with `SIGTRAP` before starting its first channel — which in practice is any config with a channel section that doesn't set `data`.

The fix just skips the copy when source and destination are the same buffer. No behaviour change on Linux; the self-copy was already a no-op there.

Verified: with this applied, `radiod` starts on macOS 26.5.2 / arm64 with a `sig_gen` config, joins its multicast groups, demodulates, and answers `metadump` and `tune` normally. Builds clean with no new warnings.

Alternative, if you'd prefer: fix it at the `config_getstring()` call site by passing `NULL` as the default and only copying on a non-NULL return, or guard inside `config_getstring()` itself. The "pass a struct field as its own default" pattern could bite again elsewhere. Happy to redo it whichever way you like.

https://claude.ai/code/session_01AqMVq98iEneqzJxwr4ij42

### DIFF
--- a/src/modes.c
+++ b/src/modes.c
@@ -491,7 +491,11 @@ int loadpreset(chan_t *chan,dictionary const *table,char const *sname){
   }
   {
     char const *data = config_getstring(table,sname,"data",chan->output.dest_string);
-    strlcpy(chan->output.dest_string,data,sizeof chan->output.dest_string);
+    // config_getstring() hands back the default when the key is absent, i.e. this very
+    // buffer. strlcpy() is undefined on overlapping objects; macOS's fortified version
+    // traps on it, killing radiod before the first channel starts.
+    if(data != chan->output.dest_string)
+      strlcpy(chan->output.dest_string,data,sizeof chan->output.dest_string);
   }
   if(!chan->use_dns || resolve_mcast(chan->output.dest_string, &chan->output.dest_socket,DEFAULT_RTP_PORT,NULL,0,2) != 0){
     // Not using DNS, or DNS resolution failed: create a IPv4 multicast address from a hash of the name


ka9q:
Fixed in a slightly different way. If I don't pass chan->output.dest_string to config_getstring() in the first place, then it can't return that pointer and strlcpy won't try to copy it.

PR #255, opened 2026-09-04
make install: load firmware into an RX888 that was already plugged in

rrobinett:
Since firmware loading moved out of radiod, an RX888 gets its firmware only when udev sees an "add" event for the bare FX3 bootloader (04b4:00f3) and starts rx888_boot.service.  On a fresh machine the radio enumerated at boot, long before 'make install' created 70-rx888-boot.rules, so no "add" event is ever seen: the radio stays in bootloader mode until it is replugged or power-cycled, and radiod restarts every 5 seconds with "rx888_usb_init() failed".  A plain 'udevadm trigger' does not help, as it synthesizes "change" events and the rule matches "add" only.

At the end of the top-level install, replay the "add" event for any such device.  This has to run after the subdirectory loop rather than from rules/Makefile because rx888_boot.service reads FIRMWARE from /etc/radio/rx888_bootfile.conf, which config/ installs after rules/.  It is skipped when DESTDIR is set or udevadm is absent, matches only unprogrammed RX888s, and never fails the install.

Seen on a greenfield Debian 13 install at N8UR with 2026.08.20.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,15 @@ install:
 	for d in $(SUBDIRS); do \
 		$(MAKE) -C $$d install DESTDIR=$(DESTDIR) || exit $$?; \
 	done
+	# An RX888 that was already plugged in enumerated as the bare FX3 bootloader (04b4:00f3) before
+	# 70-rx888-boot.rules existed, so udev never saw an "add" event for it and rx888_boot.service
+	# was never started: radiod loops on "rx888_usb_init() failed" until the radio is power-cycled.
+	# Now that the loader, its service, the rules and /etc/radio/rx888_bootfile.conf are all in
+	# place, replay the "add" event for any such device so it gets its firmware without a replug.
+	@if [ -z "$(DESTDIR)" ] && command -v udevadm >/dev/null; then \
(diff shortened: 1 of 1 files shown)


ka9q:
Looks like it should be safe to run at make install, as it doesn't seem to do anything unless there actually is an unloaded rx888 on the bus.

ka9q:
OK I put the udev trigger command in the top level Makefile but I conditioned it on ENABLE_RX888 being 1, which it is by default. I also added the command to the post-install script for the ka9q-radio-rx888 package.

I tested it with these steps:
1. apt remove ka9q-radio-rx888 ka9q-radio
2. unplug and replug rx888
3. verify that the device is waiting in DFU (bootloader) mode since the rx888_boot daemon was no longer waiting for it
4. apt install ka9q-radio-rx888 ka9q-radio

The package postinst instructions automatically issued the udev trigger command. udev invoked rx888_boot to load the device, and it dropped off and reappeared as a loaded device (00f1), then udev saw it and invoked radiod and everything was running again. No manual intervention at all.

joecupano:
Thank you !

PR #258, opened 2026-09-11
rtlsdr: return the input frequency from true_freq() when out of PLL range

mrroyh:
true_freq() models the R820T's fractional-N PLL and only produces a result when a divider brings the frequency into the VCO's range; below ~27.656MHz no divider fits. It previously returned 0 in that case, which set_correct_freq() fed straight into frontend->frequency, zeroing any tuning path that doesn't go through the PLL — direct sampling, and RTL-SDR V4's built-in upconverter, which tunes via the normal path and never sets direct_sampling.

Return freq_hz unchanged instead of 0 when no divider is found. The fix is now self-contained in true_freq(); set_correct_freq() needs no special-casing of its return value.

### DIFF
--- a/src/rtlsdr.c
+++ b/src/rtlsdr.c
@@ -463,7 +463,7 @@ static double true_freq(uint64_t freq_hz){
       break;
   }
   if(div_num > MAX_DIV)
-    return 0; // Frequency out of range
+    return freq_hz; // Not modeled by this PLL (direct sampling, V4 upconverter, etc); pass through uncorrected
 
   // PLL programming bits: Nint in upper 16 bits, Nfract in lower 16 bits
   // Freq steps are pll_ref / 2^(16 + div_num) Hz


ka9q:
I can do this, but are you sure it's the right thing to do? Both 0.0 Hz and the requested frequency are lies when the PLL is not tuned to either one. Perhaps I should just leave the tuner unchanged at its previous frequency?

mrroyh:
For plain direct sampling, the RTL2832U's own DDC is doing the tuning, at resolution far finer than the R820T's PLL steps ever were — there's no synthesizer quantization to correct for in the first place, so freq_hz is about as "true" as this librtlsdr's API can express.

Leaving frontend->frequency unchanged is actively worse, not just equally approximate: channels compute their software-NCO offset from it (it's published as FIRST_LO_FREQUENCY too). On the first tune into HF/direct-sampling mode there's no valid "previous" to fall back to — it'd sit at whatever it was initialized to, and every channel would come out mistuned by that whole delta.

ka9q:
I'm confused, mainly because I've done little with the rtlsdr and don't understand its internal structure. Can you point me to a description? I don't understand where the various functions are done. I know it has a direct sampling mode but I thought that was a software bypass.

mrroyh:
Here's what's actually happening, working from the driver source — direct sampling isn't just a different code path, it's a real hardware reconfiguration.

Turning it on does two things: `rtlsdr_set_direct_sampling()` calls `dev->tuner->exit(dev)`, which powers the R820T off over I2C, and it rewrites the RTL2832U's own ADC input-mux register (0x08) to read from the I-ADC/Q-ADC pin instead of the tuner's output. So the antenna signal is physically rerouted — it never reaches the tuner at all, straight into the ADC.

Once in that mode, picking a frequency really is pure software, as you'd expect: `rtlsdr_set_if_freq()` just writes a number into three registers on the RTL2832U itself, shifting its own digital downconverter. No PLL, no tuner chip — just ~6.9Hz steps, evenly spaced across the whole range (28.8MHz/2^22), unlike the R820T's PLL, where the step size varies by band.

That's why I think true_freq() returning 0 here is a real bug rather than a harmless default: it's trying to work out what the R820T's PLL landed on, but the R820T is powered off in this mode — there's no PLL result to find, because the thing it's modeling isn't running. The function has no way to know that, though; it runs the R820T math anyway, finds no match (direct-sampling frequencies are far below anything the R820T could tune to), and falls through to 0. And it's directly visible in control: radio_status.c publishes frontend->frequency under FIRST_LO_FREQUENCY, and control.c prints that back out labeled exactly "First LO." So on HF over direct sampling, control shows "First LO: 0.000 Hz" no matter what frequency is actually set. It's not just cosmetic, either: radio.c computes each channel's software IF shift as f - Frontend.frequency, so with that stuck at 0 the shift comes out to the full RF frequency itself — millions of Hz, way outside what the front end actually captures. That matches what I'm seeing: tuning to an HF frequency in direct sampling mode gets no signal at all.

ka9q:
I would have thought that in direct sampling mode it would do exactly that so radiod would see a "first LO" of 0 Hz. But apparently even in that mode it does its own downconversion, so I wouldn't call it "direct sampling". But I'll take your word for it since I don't use the RTL-SDR.

PR #259, opened 2026-09-12
add rtp=no to let a channel skip RTP framing for direct unicast to plain UDP tools

mrroyh:
no framing also means no loss detection, which matters most on iq-preset channels (demod=linear, stereo=yes) since a dropped datagram can permanently swap i/q with nothing left to track the pair boundary. fine for casual audio, riskier for anything doing coherent processing downstream.

playback (match rate/channels/port to the channel's rtp=no config):

  `ffmpeg -f s16le -ar <rate> -ac <channels> -i udp://@:<port> -f alsa default`

or via socat, if you'd rather not depend on ffmpeg's own udp handling:

  `socat -u UDP-LISTEN:<port>,reuseaddr - | ffmpeg -f s16le -ar <rate> -ac <channels> -i - -f alsa default`

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -76,7 +76,10 @@ int send_output(chan_t * restrict const chan, float const * restrict buffer, int
     chan->output.silent = false;
 
     uint8_t packet[PKTSIZE];
-    uint8_t * const dp = (uint8_t *)hton_rtp(packet,&rtp); // First byte after RTP header to be written
+    // rtp=no: send the bare payload with no RTP header at all (unicast point-to-point only --
+    // see loadpreset() in modes.c). RTP sequence/timestamp bookkeeping below still runs
+    // unchanged so status/metadata reporting is unaffected.
+    uint8_t * const dp = chan->output.no_rtp ? packet : (uint8_t *)hton_rtp(packet,&rtp);
     int chunk = frames;
     float const *buf = buffer;
 
--- a/src/modes.c
+++ b/src/modes.c
@@ -162,6 +162,7 @@ char const *Channel_keys[] = {
   "raster8",
   "raster9",
   "recovery-rate",
+  "rtp",
   "samprate",
   "shift",
   "snr-squelch",
@@ -470,6 +471,10 @@ int loadpreset(chan_t *chan,dictionary const *table,char const *sname){
     if(data != NULL)
       strlcpy(chan->output.dest_string,data,sizeof chan->output.dest_string);
   }
+  // rtp=no strips the RTP header from the wire; rtp state (ssrc/seq/timestamp) still
+  // updates normally below for status reporting. Inverted-sense self-referential default
+  // (same idiom as use_dns above) so an absent key leaves the current value unchanged.
+  chan->output.no_rtp = !config_getboolean(table,sname,"rtp",!chan->output.no_rtp);
   if(!chan->use_dns || resolve_mcast(chan->output.dest_string, &chan->output.dest_socket,DEFAULT_RTP_PORT,NULL,0,2) != 0){
     // Not using DNS, or DNS resolution failed: create a IPv4 multicast address from a hash of the name
     struct sockaddr_in *sin = (struct sockaddr_in *)&chan->output.dest_socket;
@@ -477,6 +482,21 @@ int loadpreset(chan_t *chan,dictionary const *table,char const *sname){
     sin->sin_family = AF_INET;
     sin->sin_addr.s_addr = htonl(addr);
     sin->sin_port = htons(DEFAULT_RTP_PORT);
+  }
+  if(chan->output.no_rtp){
+    // rtp=no drops the RTP header entirely, so SSRC can no longer distinguish channels
+    // sharing a destination. Only safe with a true unicast, one-channel-per-destination
+    // target (dns=yes, data=host:port). Warn rather than block, in case someone really
+    // does want a single-channel multicast group this way.
+    struct sockaddr const *sa = (struct sockaddr *)&chan->output.dest_socket;
+    bool is_mcast = false;
+    if(sa->sa_family == AF_INET)
+      is_mcast = (ntohl(((struct sockaddr_in const *)sa)->sin_addr.s_addr) >> 28) == 0xe;
+    else if(sa->sa_family == AF_INET6)
+      is_mcast = ((struct sockaddr_in6 const *)sa)->sin6_addr.s6_addr[0] == 0xff;
+    if(is_mcast)
+      fprintf(stderr,"%s: warning: rtp=no with multicast destination %s -- streams sharing that group will be unrecoverable without RTP/SSRC framing\n",
+	      chan->name,chan->output.dest_string);
   }
   // --> Should ensure the channel data stream is distinct from the radiod status port !!
   // Status sent to same data stream group, different port
--- a/src/radio.h
+++ b/src/radio.h
@@ -266,6 +266,7 @@ struct channel {
     double deemph_state_right;
     uint64_t samples;
     bool pacing;           // Pace output packets
+    bool no_rtp;           // Set by rtp=no: skip RTP header on the wire (unicast point-to-point only)
     enum encoding encoding;
     float *queue;          // delayed output data for aggregation when minpacket > 0
     int queue_length;      // Size of allocation, in floats


ka9q:
I could do this but I'm wondering why you want it. Why not use 'pcmrecord --stdout' to process the rtp stream and put a sequenced stream with the proper sample count on stdout, which you can then pipe into ffmpeg or whatever.

mrroyh:
i used it mainly to listen to multiple ssb freq by passing the stream to audio detector docker container to trigger recording when speech is detected.

the incovenient with pcmrecord are to compile it inside docker with it dependency.
 
while we can use ssh to execute pcmrecord and pipe it to container, i have to create ssh key and create user for it on host system.

Or is there beter way to just pipe pcmrecord --stdout to ffmpeg in docker?

ka9q:
I know what docker is, though I haven't used it because I haven't seen 
any need for containers yet.

I'm now making debian binary packages available for installation with 
apt-get so you don't have to compile anything. Just install what you 
need. Create the file /etc/apt/sources.list.d/ka9q-sources.list with 
this line:

deb [signed-by=/usr/share/keyrings/ka9q-radio.asc arch=amd64] 
http://www.ka9q.net/debian trixie main

Change "amd64" to "arm64" if you're using a RPi. Change "trixie" to the 
name of your debian release if necessary; other supported values are 
bookworm (arm64 only), forky, resolute, noble, jammy.

Also create the file /usr/share/keyrings/ka9q-radio.asc with these 
contents. This is the public key needed to verify signatures on the 
packages at www.ka9q.net

-----BEGIN PGP PUBLIC KEY BLOCK-----

mQINBGn4MQ4BEACpepjTRb3vWmGdc0CMbAUjbAO5aLTgcnkUgA5mqI6VNlucKv0l
5nbraaD7yAWIoy6JRTRblMb1xHrUpurh6kcaIrrux+EXCCbtsipNj9ZUSjGzZNC+
xOO0hcEYQu9XEpHqFIEa/4cMqO3QO47kDnU6Yc/9X1vTVqdevVHI2kFPJQGil1IU
ILBYYcoGNWnZUqpSSpgDBeObfcS3tnwPXnvAZQqfW8vDclft31FUhsoBqUyA26oA
JEb5msyjGaP00yJz2rWSc5//DWZ+bO2RN/7+rIFUid7a2Zysgs2cSsTs/nqjX6EV
yqivWiMa3DZ9K7Vzm2QkyOoTxytPVUegiQjb6DR30w2Gb1UGL62ry+D44eVejWzr
CnQ+fQRGFBHtfURZu6wsxRWNFYCM9hPdrAlEDJkrIc/U/YJjlWe42Aa+J7jTX4Wh
4JU1gAkGfHTZWwNbwWnrxzuw86bKTw2WoWWzEx1tQ1L9h4kXPKVs8Edf5p9+BMn8
jhFAaFzmDCwTMy3XtSAZlbbSX0y86DjyOuagv5+min0txQSe0ro5EbnVDveEBmyC
WeV30KnO67JsmTaQH6bvwQfGiSXkr5EsKpkzVItETc5zAKUV45oY4ae1z5B1J+ZZ
rxwwFk33kZHH7eje9UWzxg3gSjBrMIT0gB/FJPsp2QK/eGe1nHHSQ2/UqQARAQAB
tExLQTlRIFJlcG9zaXRvcnkgKHNpZ25pbmcga2V5IGZvciBrYTlxLXJhZGlvIGRl
YmlhbiBwYWNrYWdlcykgPGthcm5Aa2E5cS5uZXQ+iQJtBBMBCABXFiEEVn+1Uyjp
uioKsQwnc4U3dQlV5W0FAmn4MQ4bFIAAAAAABAAObWFudTIsMi41KzEuMTIsMCwz
AhsDBQsJCAcCAiICBhUKCQgLAgQWAgMBAh4HAheAAAoJEHOFN3UJVeVtnp4P/3zj
eK6qc53tC+iF6X5Lqu1dCJKZO+dSbbSWmaoIUP22w5zbiKKryrt+QG1LmEP+s06N
inQGJs6ikBmpFF7P0It/1eC3o04XPQGr1BvpiagdBSjdUNOX0EHvJ3J6UGzWkrUa
BbgvbYTwHniNJD4M/fuHwYAL3SnYRiBGloy8E4biDBqTHQiVRH4oEvH7PliHQ9/A
dVr9nGRi0RGwQp/56Y5AZGYsUSAOzhN4sJ34njEWHfkpquH9qu9iPl2Blb9svphU
dz5u9fdMdg5VOK1yEaKCVejz1GGiGNNJvAqESX1A+2wWP24RiXRXwrcxl280KV8/
sP+/6OyxbR7WqS04C89cALzWK1yM/pZwP/QyO2Hdvr7spah2q7nxSiu4br8N4xxI
ukgjW1Ivc+KN+WtXnIn2uDxYFfz4r+fDg5gyAIqeCZKck64EnLphFEDHnDe0l2+u
vla1AS6tyu1Nljzo+La6z/dTE1Q0oC9fJeld64eJ+PU13HUCqps1lVUJqWrjDlp5
ZbjpPTGoMUVOD45ERZXcdE4q57ixAgZ6dxDMcvKmSlRmsx4a6/nWAefJFDm6h1dC
eCYK/syF4ER2OZ7LMOF22bDUps7X7ll+x8HEpxs56bFZU+JS5RIeWxf+yQGe/7nM
/1psJTfrmHiquGdwFnRXxTa94dKbwsjvqySBAWcB
=wUXg
-----END PGP PUBLIC KEY BLOCK-----

Once you've done this, say 'sudo apt update' and you should see 
www.ka9q.net polled for its list of packages. I've divided ka9q-radio 
into a bunch of packages so you don't have to install stuff you don't 
need. To see a summary, say

apt search ka9q-radio

As with other packages managed by apt, any needed dependencies are 
automatically fetched when you install something. So for example to 
install the ka9q-radio daemon 'radiod' with support for the rx888, say

sudo apt install ka9q-radio ka9q-radio-rx888 ka9q-radio-tools 
ka9q-radio-control ka9q-radio-monitor

Or if you're feeling brave, just install everything:

sudo apt install ka9q-radio-full

This should give you enough of a working environment to use my pcmrecord 
command to receive a multicast stream so you can pipe it to an 
application. You can also use 'tune' or 'control' to set up and 
configure the channel, or do it statically in the /etc/radio config file.

73, Phil

mrroyh:
right if the docker use debian as base, we can rebuild the image with the package source.
but some docker image use very small linux distribution like alpine.
yes we can automate the rebuild that container too but its a bit incovenient.
besided its better not to add the rtp with radiod and then use extra program to strip the rtp and output the raw data (CMIIW: i am not sure if this is the behaviour of pcmrecord --stdout)

ka9q:
Yes, that's exactly what pcmrecord does; it decodes RTP, resequences
packets out of order and emits zeroes or silence when packets are
completely missing. What it emits depends on the codec in use by radiod. If
it's one of the PCM formats (integer, floating point, mu/A-law) it emits a
.wav header with metadata, format, sample rate and channel count ahead of
the raw samples.

If it's Opus, which I always use if it's just audio I'm going to listen to,
pcmrecord encapsulates the Opus frames in an Ogg container stream, also
with appropriate metadata. (If you record to disk, you get a .opus file).
Again it resequences out of order packets (important with Opus because the
decoder is stateful) and it emits empty (silent) Ogg Opus packets when
input data is dropped. You need some sort of streaming container with Opus
because the decoder needs to know where the frame boundaries are, and a raw
byte stream doesn't provide this. I regularly run commands like this:

ssh remote.system.com 'pcmrecord --stdout --ssrc 1234 mcastgroup.local' |
play -t opus -

so I can listen to a remote receiver, although latency is reduced if I
configure radiod to send RTP/UDP/IP directly to a local unicast IP address,
eg, over a Wireguard tunnel. I'd like to set up a more general purpose
routed/bridged multicast forwarding system so all my tools will work
remotely as-is but I haven't done that yet.

The RTP header also includes the 32-bit SSRC field, which I use as a stream
identifier to allow several streams to use the same multicast output group;
as you correctly note, without RTP you'd be limited to one stream per
multicast group.

I think you gain very little and lose a lot by eliminating the RTP header.

Phil

joecupano:
mrroyh - Why couldn't your container just consume the multicast stream?

mrroyh:
this is my use case and i am sorry if this is very unique one:
i have a LoG antenna -> rtlsdr -> small armbian tv box (radiod) -> tailscale -> old pc (ubuntu -> speech detector container).
the incovenient is that i have to add extra porgram like pcrmrecord to strip the rtp than feed it to ffmpeg, while ssb listening in my own opinion does not need rtp's function.

@joecupano i cannot figure out how to consume multicast in container without installing pcmrecord or compiling it

ka9q:
Why not just install pcmrecord? It’s designed to do just what you want.

mrroyh:
fair points on both - pcmrecord already resequences/silences on loss, and the ssrc multiplexing is real value i'd lose for the ssb/docker case.

different angle: gr-rtp (github.com/fventuri/gr-rtp) is a whole out-of-tree gnuradio block that exists only because gnuradio's own udp source block can't read ka9q-radio's rtp, and pcmrecord's wav/ogg-wrapped stdout isn't native gnuradio format either. rtp=no lets a flowgraph point straight at radiod with the stock block, no custom parsing needed.

another one: an esp32 (or an arduino-class board with wifi) can join the lan, open a udp socket, and push the bytes straight to an i2s dac/amp into a small speaker - a tiny standalone network radio, e.g. pointed at a wwv channel. pcmrecord isn't installable there at all, it's bare-metal firmware with no os, so "just install pcmrecord" doesn't apply even in principle.

not saying rtp is worthless, just that there's a real ecosystem already building around ka9q-radio (gnuradio users, gr-rtp) plus a whole class of hardware where a bare stream isn't just more convenient, it's the only thing that's even possible. and since rtp=yes stays default, cost is ~zero if nobody else ends up using it.

